Issue #9, opened 2026-06-29
Houdini shelf startup can deadlock when mcp.health calls hou.* from hwebserver worker thread


mrcallous:
## Environment

- fxhoudinimcp: 1.3.0
- Houdini: 21.0.512
- OS: Windows 11 x64

## What happened

Clicking the Houdini shelf tool `fxhoudinimcp::toggle` failed with:

```text
RuntimeError: hwebserver did not answer mcp.health on port 8100

```
Port 8100 was listening, but requests to /api timed out.
Likely cause
startup.start() calls _wait_for_current_process_health() from the shelf/startup path, which sends a blocking HTTP request back into Houdini.
The registered mcp.health endpoint calls hou.*:
```python
return {
    "status": "ok",
    "houdini_version": hou.applicationVersionString(),
    "hip_file": hou.hipFile.name(),
    "pid": os.getpid(),
}
```
In my environment this appears to deadlock or stall because hwebserver serves the request from a worker thread while the shelf/startup path is waiting synchronously.
Workaround verified
Changing mcp.health to avoid hou.* calls fixed startup:
```python
@hwebserver.apiFunction(namespace="mcp")
def health(request):
    return {
        "status": "ok",
        "pid": os.getpid(),
    }
```
After restarting Houdini:
```text
[fxhoudinimcp] Server ready on port 8100 (Houdini unknown, pid 13340)
```
And:
```python
{"status": "ok", "pid": 13340}
```
Suggested fix
Keep mcp.health free of hou.* calls, or split it into:
- a lightweight startup/liveness endpoint returning only status and pid
- a richer session-info endpoint that uses the normal main-thread dispatch path

guilhermecasagrandi:
## Environment

- fxhoudinimcp: 1.3.0
- Houdini: 21.0.700
- OS: Windows 11 x64

## Problem (maybe similar to above's)

Here it's not causing a deadlock, but hwebserver doesn't seems to start.
```text
Auto-start failed: hwebserver did not answer mcp.health on port 8200
```

Waiting for suggestions or a fix....

agentxray00:
[...]

PR #13, opened 2026-07-24
gate: make the gate un-escalatable — ADR-0007 Phases 1-3

husman2012:
Fork half of Homedini **ADR-0007 Phase 1 — the Act-safe gate floor**. Pairs with Homedini commit `f8692b5`.

## What
`_gated_dispatch` now hard-DENYs `scene.new_scene` / `scene.load_scene` / `cache.clear_cache` **before** `policy.decide`, regardless of mode — the floor below every mode including `ACT_SAFE`/`TRUSTED`. Keyed on the dispatch command string because `decide()` is tool-name-blind; consults `homedini.dcc.mcp_gate.never_list.is_irreversible_command` (pure-logic, shared with the panel's CLI fence).

## Why
The Homedini in-Houdini agent panel puts the gate in `trusted` at spawn, and `decide` is `TRUSTED + * → ALLOW`. Cross-vendor review found an agent under trusted runs `execute_python` → `hou.hipFile.newFile()` and wipes the scene, walking around any command-name list (finding **B1**). The Homedini side adds an `ACT_SAFE` mode denying CODE_EXEC; this fork change is the named-irreversible-command floor. Sound only together.

## Tests
- Fixture updated to provide `never_list` — the new import broke 4 existing ALLOW-path tests, caught off-DCC.
- +4 never-list DENY tests (DENY fires before `decide`); +2 ACT_SAFE mode round-trip tests through the real gate handlers. 16/16 off-DCC.

## Remaining rung
`gate_hython_smoke.py` (operator's Houdini; handler-reloading restart).


healkeiser:
Thanks for putting this up, and for the level of rigour in it. The adversarial-review notes, the ADR references, and comments like the one in `gate/__init__.py` about resolving `install_gate` at call time so `importlib.reload` cannot leave a stale binding installed are careful work. My problem with this PR is fit, not quality.

## The blocker

On this branch, in Houdini 22.0.368:

```
[fxhoudinimcp] Loaded 29/29 handler modules (221 commands registered)
RESULT: start() FAILED -> ModuleNotFoundError: No module named 'homedini'
```

`startup.py` calls `install_gate()` unconditionally, and the gate imports `homedini.dcc.mcp_gate.*` across 18 files. `homedini` is not in this repo and is not a declared dependency. Your own comment states the intent:

```python
# Fail-hard on import: if homedini.dcc.mcp_gate is missing, we never start.
```

That is the right call in your fork. A gate that silently no-ops when its core is absent would be worse than no gate. But merged here it means the plugin refuses to start for everyone who does not have the Homedini package, which is everyone using this repo. That alone rules out merging as-is, independent of anything else.

It is also self-described as "sound only together" with Homedini commit `f8692b5`. Landing half of a security invariant upstream would advertise a protection that is not actually present.

## The description and the diff are different sizes

The body describes one focused change: hard-DENY three irreversible commands before `policy.decide`, plus six tests. The PR is **76,435 additions across 165 files in 76 commits**, with subjects referencing `PP12-*` and `HMDNI-*` tickets and one titled "docs: reposition fork". It also adds 42 commands (221 against 179). That reads as a fork's accumulated history pointed at upstream rather than the change the description covers, which makes it very hard to review as a unit.

## Three other things

**It would ship its tests to PyPI.** 34,712 of the additions live in `python/fxhoudinimcp/tests/`, inside the installed package. I built the wheel: 55 test files, 1.51 MB, going to every user. Moving them to the top-level `tests/` would fix it.

**It conflicts with `main`.** Ten files were touched by both sides, including `startup.py`, `hwebserver_app.py`, `bridge.py` and `server.py`. Those were all rewritten for v2.0.0, which landed after this PR opened, so the conflicts are substantive rather than textual. Relevant overlaps: `mcp.health` no longer calls `hou.*` at all (it was deadlocking against the startup readiness poll, #9), the server now binds `127.0.0.1`, and the dispatcher's responses are JSON-encoded locally so an unserialisable value cannot escape as a bare 500 (#15).

**The suite does not pass outside your fork**: 26 failed, 110 passed, 52 collection errors, since 10 test files also import `homedini`.

## What I would like to take from it

Two of your fixes are good and one is better than what I shipped, so I want to be specific rather than hand-wave.

**Your off-main-thread health wait is better than mine.** You run the readiness poll on a daemon thread in graphical sessions so the UI never stalls, noting it does "ONLY urllib HTTP + os.getpid() -- NO hou.\* calls". That is the same constraint I arrived at fixing #9, resolved better: I only widened the main-thread stall budget from 3s to 10s. Yours removes the stall. It is safe precisely because `mcp.health` is now HOM-free, so the two compose. With your agreement I would like to adopt that idea on main with credit to you, rather than leave it sitting in a PR that cannot merge.

**Port auto-increment for concurrent instances** addresses a real gap: main currently just errors with "port owned by another Houdini process". One open question before I would take it, though. The MCP client is configured with a fixed `HOUDINI_PORT`, so a server that relocates to 8101 becomes undiscoverable unless the client also scans. How does the Homedini side handle that?

I also noticed you independently hit the false "auto-start failed" from #9/#10 and reached the same diagnosis in `widen cold-start health probe 3s->15s`.

## Suggested path

The gate belongs in your fork; it cannot work upstream while it hard-depends on a package upstream does not have, and I would not want to weaken it into a no-op just to make it importable here. What I think would land quickly is a handful of small PRs against current `main`, each with no `homedini` import:

1. The off-main-thread health wait.
2. Port auto-increment, assuming there is an answer for client discovery.
3. Any of the new handler families you want upstream (KineFX, export/VAT, USD export, spatial and temporal reasoning), one family per PR, with tests under the top-level `tests/`.

Each of those is reviewable on its own and none require the gate. I am happy to review them promptly. Leaving this PR open for now rather than closing it, so you can decide how you want to split it.

husman2012:
Hi Healkeiser,

We can close this (not sure if I'm able to do that), this was actually an
accidental PR, I was trying to keep things in my own fork, didn't realize
it had opened it on your main repo. Apologies.

Thanks,

Hamza

healkeiser:
Understood, closing as opened by mistake. No harm done.

For the record in case it is ever useful: your off-main-thread health wait is now on `main` and shipped in v2.0.1, credited to you. It was a better fix than the one I had released, since I had only widened the main-thread stall budget rather than removing the stall.

The rest of my earlier notes stand if you ever want to send pieces upstream deliberately: the port auto-increment (I would still want an answer for how the client discovers a relocated port), and any of the new handler families, one per PR against current `main`, with tests under the top-level `tests/`. The gate itself is a fork concern and I would not ask you to weaken it to make it importable here.

Note `main` has moved since this branch: v2.0.0 rewrote `startup.py`, `hwebserver_app.py`, `bridge.py` and `server.py`, and v2.1.0 ships the Houdini plugin inside the Python package. Worth rebasing on that before extracting anything.

PR #14, opened 2026-07-25
Houdini 22 support: python3.13libs, USD 0.26.5, PDG deadlock, reconnect, hou.pdg removal

voxeldance:
Houdini 22 support, plus two bugs that cost a session when they hit.

Tested against **Houdini FX 22.0.368** (Python 3.13, USD 0.26.5) on macOS by sweeping all 179 registered commands through the live plugin in a GUI session. Every command has at least one run against a valid target. The existing 102 unit tests still pass.

Everything here is verified on 22 specifically — I don't have 20.5/21 to test against, so I've kept the changes backward compatible rather than claiming they're validated there.

Each commit is independent, so feel free to take a subset.

---

### 1. `python3.13libs` so the plugin loads at all

Houdini loads the `pythonX.Ylibs` directory matching its own bundled Python. Houdini 22 ships Python 3.13 and only 3.9-3.11 are present, so nothing matches and `uiready.py` never runs. The failure is silent: no error, no log line, the server simply never comes up on port 8100 and the MCP client reports a connection failure that looks client-side.

`uiready.py` is identical across the shipped version directories, so this is a copy of `python3.11libs`.

### 2. `Sdf.PrimSpec.nameChildren` iteration under USD 0.26.5

Houdini 22 ships USD 0.26.5, where iterating a `PrimSpec`'s `nameChildren` yields **`PrimSpec` objects**; earlier USD yielded name strings. Feeding those to `Sdf.Path.AppendChild` raises:

```
ArgumentError: Python argument types in Path.AppendChild(Path, PrimSpec)
did not match C++ signature: AppendChild(SdfPath {lvalue}, TfToken)
```

This breaks `lops.inspect_usd_layer` and the fallback path of `lops.get_last_modified_prims`.

Iterating `.keys()` yields `str` on both old and new USD, so the fix is backward compatible.

### 3. Blocking PDG calls can deadlock Houdini

**This one hard-freezes the application and requires a force-quit.** It cost me three sessions before I isolated it.

`generateStaticWorkItems(True)` and `cookWorkItems(block=True)` run on the main thread, because `dispatcher.dispatch` marshals every `hou.*` call through `hdefereval.executeInMainThreadWithResult()`. When the target cannot produce work items, the blocking call waits on a cook that never starts — and the main thread is precisely what would advance that cook. The UI stops responding permanently; the process stays alive but has to be killed.

Reproduction on a clean scene:

```
tops.generate_static_items(node_path="/obj/topnet1")   # fresh topnet -> hangs Houdini
```

A freshly created topnet triggers it, because it contains only a `localscheduler` and a TOP network generates through its display node. So "make a topnet, ask for work items" — plausibly a first thing to try — takes out the session.

The guard checks the PDG layer rather than matching type names: a scheduler's PDG object has no `nodeType` attribute, while Processor/Partitioner/Mapper nodes do (`pdg.nodeType` has no `Scheduler` member). Populated networks still generate and cook unchanged; the bad case now returns in ~0.05s with a message naming the scheduler and suggesting adding a TOP node.

Not version-specific — worth having anywhere.

### 4. Reconnect after Houdini restarts

Houdini closes its keep-alive connections on exit, so the first request afterwards reuses a dead socket and httpx raises `RemoteProtocolError`. Nothing caught it, so the server reported *"Server disconnected without sending a response"* and stayed unusable until the MCP client itself was restarted — even though a freshly started server connected to the same Houdini fine.

That asymmetry is the giveaway: restarting Houdini during a session shouldn't require restarting the MCP client.

Fixed by retrying once on a fresh connection pool. Verified by restarting Houdini underneath a live server: it reconnected to the new pid and the next tool call succeeded.

One note while routing `health_check` through the shared helper — `timeout=None` in httpx means *wait forever* rather than "use the client default", so the helper falls back to `self.timeout` to preserve the previous behaviour.

### 5. `hou.pdg` was removed in Houdini 22

`hasattr(hou, "pdg")` is `False` on 22, so the fallback in `_get_graph_context` could never resolve anything: it raised `AttributeError`, which the `except` swallowed, and every lookup ended in *"Cannot find PDG graph context"*.

The visible effect is that `tops.pause_top_cook` and `tops.cancel_top_cook` fail whenever handed a TOP **network**, even though `generate_static_items`, `cook_top_node` and `get_work_item_states` all accept one:

```
tops.pause_top_cook(node_path="/obj/topnet1")         # ValueError: Cannot find PDG graph context
tops.pause_top_cook(node_path="/obj/topnet1/wedge1")  # works
```

A TOP network has no PDG node of its own and cooks through the node it displays, so the context is taken from there. The legacy `hou.pdg` lookup is kept after it for older versions.

---

### DIFF
--- a/houdini/python3.13libs/uiready.py
+++ b/houdini/python3.13libs/uiready.py
@@ -1,0 +1,18 @@
+"""Auto-start FXHoudini-MCP server when Houdini's UI is ready.
+
+This script is sourced by Houdini once at startup, after the UI is
+initialised.  Unlike scripts/456.py it stacks correctly with other
+packages that also define a uiready.py.
+
+Set FXHOUDINIMCP_AUTOSTART=0 to disable auto-start.
+"""
+
+import os
+
+if os.environ.get("FXHOUDINIMCP_AUTOSTART", "1") == "1":
+    try:
+        import fxhoudinimcp_server.startup
+
+        fxhoudinimcp_server.startup.ensure_running()
+    except Exception as e:
+        print(f"[fxhoudinimcp] Auto-start failed: {e}")
--- a/houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
+++ b/houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
@@ -423,13 +423,13 @@ def _get_last_modified_prims(*, node_path: str) -> dict[str, Any]:
         spec = layer.GetPrimAtPath(path)
         if spec:
             authored.append(str(path))
-            for child_name in spec.nameChildren:
+            for child_name in spec.nameChildren.keys():
                 _walk(path.AppendChild(child_name))
 
     root = Sdf.Path.absoluteRootPath
     root_spec = layer.GetPrimAtPath(root)
     if root_spec:
-        for child_name in root_spec.nameChildren:
+        for child_name in root_spec.nameChildren.keys():
             _walk(root.AppendChild(child_name))
 
     return {
@@ -764,13 +764,13 @@ def _inspect_usd_layer(
         if spec:
             authored_prims.append(str(path))
             if hasattr(spec, "nameChildren"):
-                for child_name in spec.nameChildren:
+                for child_name in spec.nameChildren.keys():
                     _walk_layer(path.AppendChild(child_name))
 
     root = Sdf.Path.absoluteRootPath
     root_spec = layer.GetPrimAtPath(root)
     if root_spec and hasattr(root_spec, "nameChildren"):
-        for child_name in root_spec.nameChildren:
+        for child_name in root_spec.nameChildren.keys():
             _walk_layer(root.AppendChild(child_name))
 
     # Sublayers
--- a/houdini/scripts/python/fxhoudinimcp_server/handlers/top_handlers.py
+++ b/houdini/scripts/python/fxhoudinimcp_server/handlers/top_handlers.py
@@ -91,6 +91,17 @@ def _get_graph_context(node: hou.Node):
     parent = node
     while parent is not None:
         if parent.type().category().name() == "TopNet" or parent.type().name() == "topnet":
+            # A TOP network has no PDG node of its own -- it cooks through the
+            # node it displays -- so borrow the context from there. Houdini 22
+            # removed hou.pdg entirely, so the legacy lookup below cannot
+            # resolve anything there and this is the path that works.
+            display = parent.displayNode() if hasattr(parent, "displayNode") else None
+            if display is not None:
+                display_pdg = display.getPDGNode()
+                if display_pdg is not None and display_pdg.context is not None:
+                    return display_pdg.context
+
+            # Legacy path for Houdini versions that still expose hou.pdg.
             try:
                 contexts = hou.pdg.GraphContext.contexts()
                 for ctx in contexts:
--- a/houdini/scripts/python/fxhoudinimcp_server/handlers/top_handlers.py
+++ b/houdini/scripts/python/fxhoudinimcp_server/handlers/top_handlers.py
@@ -39,6 +39,47 @@ def _get_pdg_node(node: hou.Node):
     return pdg_node
 
 
+def _assert_has_generating_node(node: hou.Node) -> None:
+    """Raise unless *node* resolves to a TOP node that produces work items.
+
+    Blocking PDG calls (generateStaticWorkItems(True), cookWorkItems(block=True))
+    run on the main thread here, because the dispatcher marshals every hou.*
+    call through hdefereval. When the target cannot produce work items the
+    blocking call waits on a cook that never starts, and since the main thread
+    is what would advance that cook, Houdini's UI deadlocks and the app has to
+    be force-quit. A freshly created topnet hits this: it contains only a
+    localscheduler, so it is the common case rather than an exotic one.
+
+    Schedulers are the tell -- their PDG object has no ``nodeType``, while
+    Processor/Partitioner/Mapper nodes do.
+    """
+    target = node
+    if node.getPDGNode() is None and hasattr(node, "displayNode"):
+        # TOP networks have no PDG node of their own; they generate through
+        # whichever child is displayed.
+        target = node.displayNode()
+
+    if target is None:
+        raise ValueError(
+            f"TOP network {node.path()} is empty, so there is nothing to "
+            "generate. Add a TOP node (for example wedge or filepattern) first."
+        )
+
+    pdg_node = target.getPDGNode()
+    if pdg_node is None:
+        raise ValueError(
+            f"Node {target.path()} has no PDG node, so it cannot generate "
+            "work items. It may not be a TOP node."
+        )
+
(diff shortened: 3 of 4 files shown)


voxeldance:
Pushed a fifth commit: **hou.pdg was removed in Houdini 22**.

`hasattr(hou, "pdg")` is `False` there, so the fallback in `_get_graph_context` could never resolve anything — it raised `AttributeError`, which the `except` swallowed, and every lookup ended in *"Cannot find PDG graph context"*.

The visible effect is that `tops.pause_top_cook` and `tops.cancel_top_cook` fail whenever they are handed a TOP **network**, even though `generate_static_items`, `cook_top_node` and `get_work_item_states` all accept one:

```
tops.pause_top_cook(node_path="/obj/topnet1")   # ValueError: Cannot find PDG graph context
tops.pause_top_cook(node_path="/obj/topnet1/wedge1")  # works
```

A TOP network has no PDG node of its own and cooks through the node it displays, so the context is taken from there. The legacy `hou.pdg` lookup is kept after it for older versions.

Found while closing a gap in my own testing: five commands had only ever been exercised with deliberately invalid arguments, so I went back and re-ran them against a real RBD sim, a real pyro sim, and a topnet with real work items. Four turned out to be my bad arguments; this one was a genuine bug. All 179 registered commands now have at least one run against a valid target, and the 102 unit tests still pass.

voxeldance:
Narrowed the scope of the first commit: it now adds **only `python3.13libs`** (Houdini 22's Python), not 3.12/3.14.

I had included those to cover the Houdini 21/23 era, but I only have Houdini 22 to test against, so shipping directories I can't verify was speculation rather than a fix. Force-pushed the rebuilt branch; no reviews were affected.

Everything in the PR is now verified against 22.0.368 specifically. The changes stay backward compatible, but I'm not claiming they're validated on 20.5/21.

healkeiser:
Reviewed by reproducing each claim independently on Houdini FX 22.0.368 (Windows 11), first on `main` to confirm the bug, then on this branch to confirm the fix. Four of the five hold up exactly as you describe. Merging.

Thank you for the level of detail - "each commit is independent, so feel free to take a subset" plus a repro per claim made this genuinely quick to verify.

### 1. `python3.13libs` - confirmed

Houdini 22 ships Python 3.13 and only 3.9–3.11 were present. Also landed in #16; the file is byte-identical, so expect a trivial add/add when the second of the two merges.

### 2. USD `nameChildren` - confirmed, and it can fail more quietly than reported

```
USD version: (0, 26, 5)
list(spec.nameChildren)        -> ['PrimSpec', 'PrimSpec']
list(spec.nameChildren.keys()) -> ['str', 'str']

Path.AppendChild(PrimSpec) -> ArgumentError ... did not match C++ signature
Path.AppendChild(str)      -> OK
```

Worth adding to the record: in my repro `lops.inspect_usd_layer` did not raise, it returned `authored_prims: []` on `main` versus `['/HoudiniLayerInfo']` on this branch. So depending on the layer's shape this surfaces as silently missing prims rather than an `ArgumentError`, which is arguably the nastier failure. Your fix covers both, and `.keys()` returning `str` on old and new USD makes it backward compatible as you say.

### 3. PDG deadlock - confirmed, and worse than reported headless

On `main`, `tops.generate_static_items` against a fresh topnet **segfaults hython** rather than hanging:

```
children: ['localscheduler']
25044: Fatal error: Segmentation fault
```

On this branch it returns in 0.00s with a message that names the offender, and a populated network still generates normally. Every fact behind your heuristic checks out on 22.0.368:

```
pdg.nodeType members: ['Mapper', 'Partitioner', 'Processor', 'name', 'value']
pdg.nodeType has 'Scheduler': False
fresh topnet -> getPDGNode() None, displayNode 'localscheduler' -> LocalScheduler, hasattr nodeType False
after adding wedge -> Wedge, hasattr nodeType True
```

Checking the PDG layer rather than matching type names was the right call. This is the most valuable commit in the PR - "make a topnet, ask for work items" really is a plausible first thing to try.

### 4. Reconnect - I could not reproduce the trigger, and the fix is too narrow for the gap that is real

On httpx 0.28.1 / httpcore 1.0.9 I could not get a stale pooled connection to surface as `RemoteProtocolError`. I tried both shapes: closing the socket while idle, and closing it after receiving the request bytes but before responding (the race that would produce "Server disconnected without sending a response"). httpcore transparently retried on a fresh connection both times and the request succeeded - the simulated server logged the extra attempt.

Not disputing what you saw; you quoted a specific error string, and this may well be version-dependent. But the underlying gap is real and wider than `RemoteProtocolError`. `execute` only wraps `ConnectError`, `HTTPStatusError` and `TimeoutException`, so any other transport failure leaks raw. Provoking a protocol violation on current httpx gives:

```
health_check: httpx.ReadError:      <- note the empty message
execute:      httpx.ReadError:      <- same
```

An unwrapped httpx exception with an empty string is exactly the unhelpful-diagnostic problem you are describing, reached by a different exception. Since `httpx.TransportError` is the common base for `ConnectError`, `ReadError`, `WriteError`, `RemoteProtocolError` and the timeout family, catching that covers the whole class:

```python
except httpx.TransportError as e:   # after the specific TimeoutException branch
```

No need for you to do anything - I am taking this on in #16 on top of your `_post` helper, so the retry stays and the wrapping gets broader. Keeping your commit as-is; it is strictly an improvement either way.

Your `timeout=None` aside is also correct and worth having in a comment: `httpx.Timeout(None).connect is None`, i.e. wait forever, not "use the client default".

### 5. `hou.pdg` removed - confirmed

`hasattr(hou, "pdg")` is `False` on 22.0.368, so the legacy lookup could never resolve anything. The asymmetry you reported reproduces precisely on `main`:

```
tops.pause_top_cook  on TOP network -> error "Cannot find PDG graph context"
tops.pause_top_cook  on TOP node    -> success
```

Both succeed on this branch.

---

One heads-up on what changed underneath you in #16: `mcp.health` no longer calls `hou.*` at all (it was deadlocking against the startup readiness poll), the plugin now binds `127.0.0.1` instead of the any-address, and `hwebserver_app` encodes its own JSON so an unserialisable value cannot escape as a bare 500. None of it touches the files you changed except `bridge.py`, which I will rebase onto yours.

healkeiser:
Heads-up: rather than merging this separately and leaving a `bridge.py` conflict, I merged this branch into #16 so both land in one go. Your five commits are preserved with authorship, so this PR will close as merged when #16 does.

The only conflict was `health_check`'s except clause, resolved in favour of `httpx.TransportError` - the base class of the `ConnectError` / `RemoteProtocolError` / `TimeoutException` trio your commit lists, so the catch is a strict superset and nothing is lost. Your `_post` retry helper is kept unchanged and now sits in front of that wrapping.

One thing I changed in the tests while I was in there: the transport tests now patch `_reset_client` as well as `_get_client`. With only the latter mocked, `_post`'s retry made a real connection attempt to port 8100 - a couple of seconds per test, and it would resolve differently on a machine with Houdini actually listening.

Thanks again. The PDG guard in particular is the kind of bug that costs people a session before they work out what happened, and "fresh topnet, generate work items" is far too easy to stumble into.

Issue #15, opened 2026-07-25
get_node_info returns HTTP 500 on any node with a Ramp or Data parameter


Alsoilia:
## Summary

`get_node_info` fails with `HTTP 500` on any node that has a **Ramp** or **Data (Geometry)** parameter whose value is reported as non-default. The MCP client only sees `Houdini returned HTTP 500`, with no indication of the cause.

This hits very common grooming/hair nodes, so a whole class of scenes cannot be inspected with `get_node_info`.

## Environment

- fxhoudinimcp 1.3.0 (PyPI)
- Houdini 21.0.440, Windows 11, Python 3.11.7
- Client: Claude Code, stdio transport, bridge on `127.0.0.1:8100`

## Reproduce

Any node with a ramp parameter is enough:

```
create_node(parent_path="/obj/geo1", node_type="attribpaint")
get_node_info(node_path="/obj/geo1/attribpaint1")
```

Result:

```
Error executing tool get_node_info: Houdini returned HTTP 500
```

Confirmed failing node types in my scene:

| Node type | Offending parameters |
|---|---|
| `attribpaint` | `strokegeo`, `bakedgeo` (type `Data` → `hou.Geometry`) |
| `guidecollidevdb` | `curvemaskramp` (type `Ramp`) |
| `guideprocess` | ramp parameters |
| `hairclump::2.0` | 15 ramps (`blendremapramp`, `clumpsizeremapramp`, `tightnessremapramp`, …) |
| `groom::mh_head::21.0` | `thicknessramp`, `iris_radius_ramp` |

Working normally on the same scene: `merge`, `resample`, `mirror`, `attribdelete`, `box`, `subdivide`, `mountain::2.0`, `null` — i.e. every node whose parameters are all JSON-serializable.

## Root cause

`handlers/node_handlers.py`, lines 204–224:

```python
parms_summary.append({
    "name": parm.name(),
    "label": parm.description(),
    "value": val if not isinstance(val, (hou.Vector2, hou.Vector3, hou.Vector4)) else list(val),
    "default": default if not isinstance(default, tuple) else list(default),
    "type": parm.parmTemplate().type().name(),
})
```

`hou.Vector2/3/4` are coerced to lists, but `hou.Ramp` and `hou.Geometry` are appended raw and then fail at JSON encoding time.

Verified inside Houdini:

```python
import hou, json
n = hou.node("/obj/geo1/density")   # attribpaint
[...]

Issue #12, opened 2026-07-19
Houdini 22.0.368 UI sessions: plugin unusable (thread-local hwebserver, /api shadowed, request bodies unreadable) — five root causes + working fixes


typofreak:
## Environment
- fxhoudinimcp **1.3.0** (PyPI) + repo clone for the Houdini plugin
- Houdini **22.0.368** (Indie), Windows 11, bundled Python **3.13.10**
- MCP client: Claude Code, `.mcp.json` launching `python -m fxhoudinimcp`

## Summary
On Houdini 22.0.368 the plugin is unusable in **UI sessions**: auto-start reports `hwebserver did not answer mcp.health on port 8100`, and every API call to `/api` returns an **empty 200** response. The same code works fine in headless hython on the same build. We traced it to five separate H22 changes/bugs in `hwebserver` (now a pure-Python module at `houdini/python3.13libs/hwebserver.py` wrapping `_hwebserver`), and have the plugin fully working with the fixes below.

## Root causes found (all verified against live 22.0.368)

1. **No auto-start hook for Python 3.13.** H22 runs py3.13; the repo ships `python3.9libs` / `python3.10libs` / `python3.11libs` only, so `uiready.py` never runs.

2. **hwebserver state is now thread-local.** `hwebserver.py` line ~33: `THREAD_SERVER = threading.local()` — every thread gets its own `Server` object via `__getThreadServer()`. Registering API functions on the main thread interacts badly with Houdini's built-in server; serving from a dedicated thread (fresh `Server`) works.

3. **Houdini's built-in web API shadows the `/api` path in UI sessions.** Any POST to `/api` on **any** port (including ones opened by the plugin) is answered with an empty 200 before plugin handlers are consulted. A custom path (we used `/fxapi`) is dispatched normally. Headless hython has no built-in server, which is why `/api` works there.

4. **HTTP request bodies are unreadable in UI sessions (the big one).** Verified with minimal `urlHandler` probes in a live UI session:
   - POST with `application/x-www-form-urlencoded` body → request swallowed pre-dispatch, empty 200, handler never runs;
   - POST `multipart/form-data` → same;
   - `Request.body()` from a handler → kills the request (empty 200) and can wedge the worker (subsequent request times out);
   - POST with `application/json` body dispatches, but the body is inaccessible (`POST()` empty, `body()` fatal);
   - GET, and POST **without** a body, work fine; query strings 404 above ~2–4 KB.
[...]

Issue #11, opened 2026-07-09
package.json default path ($HOUDINI_USER_PREF_DIR/fxhoudinimcp/houdini) does not exist after following README Option A


woohahahaaa:
## Environment

- fxhoudinimcp: 1.3.0
- Houdini: 21.0.440
- OS: Windows 11 x64

## What happened

Following README **Option A (Houdini package - recommended)**:

> 1. Copy `houdini/fxhoudinimcp.json` to your Houdini packages directory
> 2. Edit the JSON file to set `FXHOUDINIMCP` to point to the `houdini` directory in this repo

The default `fxhoudinimcp.json` contains:

```json
{
    "env": [
        {
            "FXHOUDINIMCP": {
                "value": "$HOUDINI_USER_PREF_DIR/fxhoudinimcp/houdini",
                "method": "default"
            }
        }
    ],
    "path": "$FXHOUDINIMCP"
}
```

After copying this to `Documents/houdini21.0/packages/fxhoudinimcp.json`, `$HOUDINI_USER_PREF_DIR` resolves to `C:\Users\wooh\Documents\houdini21.0`, so the path becomes:

```
C:\Users\wooh\Documents\houdini21.0\fxhoudinimcp\houdini
```

But the actual install location (from `git clone` or `pip install -e .`) is:

```
C:\Users\wooh\fxhoudinimcp\houdini
```

The path **does not exist**, so Houdini silently fails to load the package. No shelf, no `uiready.py`, no `fxhoudinimcp_server` module.

## Why this is confusing

1. The default value **looks like a working path** - there is no indication it needs to be changed
2. README step 2 says "edit the JSON file to set `FXHOUDINIMCP` to point to the `houdini` directory in this repo" - but a user who just copy-pasted the file might not realize the default is wrong
3. Houdini does not print any error when a package JSON fails to resolve - it just silently skips it
4. On macOS/Linux the same issue occurs (`$HOUDINI_USER_PREF_DIR` resolves to `~/Library/Preferences/houdini/XX.X` or `~/houdiniXX.X`)

## Additional gotcha: PowerShell UTF-8 BOM

If users edit the JSON on Windows using PowerShell `Set-Content -Encoding UTF8`, a UTF-8 BOM (`EF BB BF`) is prepended. **Houdini's JSON parser does not accept BOM** and silently skips the entire package. This is particularly insidious because the file looks correct in most text editors.

## Suggested improvements

[...]

Issue #10, opened 2026-07-09
hwebserver.run() in startup.py blocks main thread in Houdini GUI mode


woohahahaaa:
## Environment

- fxhoudinimcp: 1.3.0
- Houdini: 21.0.440
- OS: Windows 11 x64

## What happened

When Houdini starts in GUI mode, `uiready.py` auto-start fails with:

```
[fxhoudinimcp] Auto-start failed: hwebserver did not answer mcp.health on port 8100
```

Clicking the **MCP Server** shelf tool (`fxhoudinimcp::toggle`) gives the same error:

```
RuntimeError: hwebserver did not answer mcp.health on port 8100
```

## Root cause

In `houdini/scripts/python/fxhoudinimcp_server/startup.py` line 96:

```python
hwebserver.run(_port, debug=False)
```

In Houdini GUI mode, `hwebserver.run()` **without `in_background=True`** is a blocking call. It blocks the main thread, so the subsequent `_wait_for_current_process_health()` call never gets a chance to execute — the HTTP request to `/api` is queued behind the blocking `run()` call and times out.

This is separate from issue #9 (where `mcp.health` deadlocks due to `hou.*` calls in a worker thread). Even with a lightweight health endpoint, the blocking `run()` call prevents startup from completing in GUI mode.

## Workaround verified

Running this in the Python Source Editor works:

```python
import hwebserver, sys
hwebserver.run(8100, debug=False, in_background=True)
sys.path.insert(0, r"C:\Users\wooh\fxhoudinimcp\houdini\scripts\python")
from fxhoudinimcp_server import handlers, hwebserver_app
print("MCP endpoints registered on port 8100")
```

After this, the MCP server is fully functional — health check passes, tools work, UI is responsive.

## Suggested fix

Pass `in_background=True` to `hwebserver.run()`:

```python
hwebserver.run(_port, debug=False, in_background=True)
```

Or detect GUI mode and only pass it there:

```python
import hou
in_bg = not hou.isUIAvailable()  # hython: blocking is fine
hwebserver.run(_port, debug=False, in_background=in_bg)
```

Note: `in_background=True` should be the default for GUI mode, since `hython` users typically want blocking behavior (they can Ctrl+C to stop).

healkeiser:
Addressed in #16, but I want to be straight with you about the diagnosis rather than quietly ship the patch: `hwebserver.run()` does not block the main thread in GUI mode.

[...]

PR #16, opened 2026-07-27
Fix #9, #10, #11, #15; add Houdini 22 support, an MCP menu, and loopback binding

healkeiser:
Fixes four reported issues, adds Houdini 22 loading, and closes a security hole nobody had filed. One commit per theme, so any of them can be taken or dropped independently.

Everything here is verified against **live Houdini 22.0.368**, headless and in a **Windows 11 GUI session**. Unit tests go from 102 to 139.

## Fixes

**#15 - `get_node_info` returned HTTP 500 on Ramp/Data parameters.** `hairclump::2.0` alone carries 15 ramps, so a whole class of grooming scenes could not be inspected. Two things made it worse than one bad field: the "skip parameters at their default" guard cannot filter ramps out (`defaultValue()` yields nothing comparable), so they always reach the response; and hwebserver encodes the return value in `_check_result`, which sits *outside* the `try/except` in `APIUrlHandler._call_api`, so the `TypeError` escaped to the C++ caller as a bare 500 with no diagnostic.

New `serialize.to_jsonable` handles the HOM types that appear in results and degrades unknown objects to a truncated `repr`, so no single exotic value can take down a call. `mcp.execute` now encodes its own JSON with a `default=` hook, which fixes this class of failure for all 179 commands rather than just one. Ramps come back as `basis`/`keys`/`values` instead of a repr string. Geometry counts use intrinsics, not `len(geo.points())`, which would materialise a tuple of point objects for every parameter of every inspected node.

`attribpaint`'s `strokegeo`/`bakedgeo` evaluate to `None` on 22.0.368, so the `hou.Geometry` half of the report does not reproduce there; it is handled anyway since it was observed on 21.0.440.

**#9 - startup deadlock.** `mcp.health` called `hou.applicationVersionString()` and `hou.hipFile.name()`, and hwebserver serves API functions from a worker thread. During startup the main thread sits inside the readiness poll, so it is not running Houdini's event loop, while HOM access from the worker needs exactly that main thread. Neither side advances. `health` is now HOM-free - version comes from the `HOUDINI_VERSION` environment variable Houdini already exports - and scene data moved to a new `mcp.session_info` that uses the normal dispatch path. `bridge.health_check()` still returns `houdini_version`, so callers are unaffected.

**#10 - please note the diagnosis was wrong.** `hwebserver.run()` does not block in GUI mode: line 1673 reads `run_in_thread = kwargs.get("in_background", isUIAvailable())`, so `in_background` already defaults to `True` there and the suggested fix is a no-op. The reported symptom was #9's deadlock. `in_background` is now passed explicitly anyway so the intent is visible and immune to the default shifting, plus a `background=` argument for headless callers, and the readiness poll goes 3s → 10s.

**#11 - package default path.** The old default *looked* like a working path, and Houdini skips an unresolvable package silently. Now an unmistakable placeholder. Also drops `"method": "default"`, which Houdini rejects (`HOUDINI_PACKAGE_VERBOSE` prints `Unsupported method value: default`) and which does not mean what it looks like - a pre-set value is overridden either way.

The README now states something the issue guessed wrong: **the PyPI wheel contains no Houdini plugin files** (`packages = ["python/fxhoudinimcp"]` - verified by building it), so a pip-only user has no `houdini/` directory and cannot follow either install option without cloning. Plus the UTF-8 BOM trap and `HOUDINI_PACKAGE_VERBOSE=1` for diagnosis.

## Houdini 22

Adds `python3.13libs/uiready.py`. Houdini 22 ships Python 3.13 and only 3.9–3.11 were present, so nothing matched and the plugin never loaded - silently. (Same fix as #14, which should still be merged for its four unrelated H22 fixes.) SideFX skipped 3.12, so 3.9/3.10/3.11/3.13 now covers 20.5 through 22.0.

## Security: the port was open to the LAN

Not previously filed. hwebserver binds the any-address by default (its `ADDRESS` setting documents `0.0.0.0`), confirmed with `netstat`. This bridge runs arbitrary Python inside Houdini via `code_handlers` with no authentication, so anyone who could reach port 8100 had the session. Now bound to `127.0.0.1`, with `FXHOUDINIMCP_BIND` to reopen it deliberately.

## MCP menu replaces the shelf

A top-level **MCP** menu (Start Server / Stop Server / Server Status), verified rendering in a live GUI session. `Start` surfaces failures through `hou.ui.displayMessage` since a failed auto-start is the case users actually hit, and `Status` reports the bind address, now the first thing to check when a client cannot connect. The shelf is removed as redundant.

## On #12

Issue #12 proposes moving the transport to `/fxapi` with a GET query-string/temp-file tunnel. That is not implemented here, because its premise does not hold:

- `hwebserver.py` is **byte-identical** between 21.0.440 and 22.0.368 (md5 `95df1069dc1acc2dfb21b0aa135b01ad`). Nothing described as an "H22 change" is one - `THREAD_SERVER = threading.local()`, the `setSettingsForPort(settings, port_name)` order, the broken module-level `registerAPIFunction`, and the decorators returning `None` are all present in 21.
- Claims 3 and 4 do not reproduce. In a live **Windows 11 / 22.0.368 GUI** session, `POST /api` with a form-encoded body returns proper JSON, reaches all 179 commands, and accepts a 20 KB body. PR #14's macOS GUI sweep of all 179 commands agrees.
- Claim 2 is real but points the other way: `run()` on a fresh thread fails with `OperationFailed: No URL handlers have been added to the server`, so registering and running on one thread - what the plugin does today - is correct.

The genuine findings in #12 are the missing `python3.13libs` (fixed here) and the loopback hardening (fixed here, and more serious than the report suggested).


healkeiser:
### Update: this now also contains PR #14

Merged `houdini22-compat` in, so a single merge lands both instead of leaving a `bridge.py` conflict for whichever went second. @voxeldance's commits are preserved with authorship, so **#14 will close as merged** when this does.

All five of its claims were verified independently on 22.0.368 first (details in the [review](https://github.com/healkeiser/fxhoudinimcp/pull/14#issuecomment-5090356183)) - reproduced on `main`, then confirmed fixed on the branch. Two findings worth repeating:

- `tops.generate_static_items` on a fresh topnet **segfaults hython** on `main`, not just hangs. Their guard turns it into a 0.00s error naming the scheduler.
- The USD `nameChildren` bug can fail silently rather than raising: `lops.inspect_usd_layer` returned `authored_prims: []` on `main` versus `['/HoudiniLayerInfo']` with the fix, in my repro.

### Three commits added since the original description

**`Keep hip_file in the connection-status payload`** - backward compatibility. Making `mcp.health` HOM-free dropped `hip_file`, which would have quietly changed the shape of `get_houdini_connection_status` for anyone reading `health["hip_file"]`. It is now backfilled from `scene.get_scene_info`, best-effort with a short timeout, and skipped entirely when health already carries the key so an older Houdini-side plugin is not second-guessed. A failure there still reports `connected: true` - this is a diagnostic whose whole point is to answer without raising.

**`Wrap every transport failure`** - the general form of #14's reconnect problem. `execute`/`health_check` only caught `ConnectError`, `HTTPStatusError` and `TimeoutException`, so anything else escaped raw:

```
httpx.ReadError:            <- the entire error, empty message
```

`httpx.TransportError` is the base of that whole family, so it covers the class rather than the members we happened to name. Notably `ReadError` is what actually escaped in testing, which naming `RemoteProtocolError` alone would not have caught.

**The merge commit** - resolves the one conflict in `health_check` in favour of `TransportError` (a strict superset of the trio it replaces) and keeps #14's retry helper in front of it.

Unit tests: 102 → 151.

PR #17, opened 2026-07-27
Hold advertised node types to a declared version range (fixes 20.5 and 22.0)

healkeiser:
The instructions handed to the assistant advertised node types that only exist in part of the supported range, and `test_every_advertised_node_type_exists` could only ever pass on Houdini 21.0. Running the integration suite across every installed build, it failed on **five of six**.

## Baseline

| Version | Before | After |
|---|---|---|
| 20.5.278 | 2 failed, 120 passed | **123 passed**, 1 skipped |
| 20.5.487 | 2 failed, 120 passed | **123 passed**, 1 skipped |
| 20.5.613 | 2 failed, 120 passed | **123 passed**, 1 skipped |
| 20.5.654 | 2 failed, 120 passed | **123 passed**, 1 skipped |
| 21.0.440 | 122 passed | **123 passed**, 1 skipped |
| 22.0.368 | 1 failed, 121 passed | **123 passed**, 1 skipped |

## The problem cut both ways

**Houdini 22 (current release) was being handed names that are not there.** The LOP `instancer` was renamed to `pointinstancer`, and `layout` was dropped outright:

```
Lop 'instancer' : 21.0 exact=True  | 22.0 exact=False, related=['pointinstancer', ...]
Lop 'layout'    : 21.0 exact=True  | 22.0 exact=False, related=[]
```

**Houdini 20.5 was being pointed at 28 nodes that did not exist yet** - most of Copernicus (`colorcorrect`, `tonemap`, `sharpen`, `defocus`, `cornerpin`, `latticedeform`, the noise/3D/pyro sets), plus `karmarendersettings`, `portallight`, `geometrylight`, `gasburn`, `deflate`, `otissolver`, `neuralpointsurface`, `ml_regressioninference` and `ropflipbook`. Identical on all four 20.5 builds, so this is the 20.5→21 feature gap, not an early-build quirk.

Either way the assistant is being told to create a node that does not exist, which is precisely what this test exists to prevent.

## Approach

Names carry an inline range where they need one, and the test enforces each name against its declared range:

```
*   Instancing: instancer (20.5-21.0), pointinstancer (22.0+), modifypointinstances, ...
*   Color: colorcorrect (21.0+), hsv, gamma, bright, contrast, invert, tonemap (21.0+)
```

The markdown stays the single source of truth. A version table living in the test would drift from the text the assistant actually reads, and the assistant needs this information at least as much as the test does - so the annotation serves both, with the syntax explained at the top of the section and a pointer to `get_scene_info` for the running version.

32 names annotated. Every range was verified against all six builds: the test also reports any name that turns up *outside* its declared range, and after annotating, none do.

`layout` now says plainly that it is gone in 22.0 and names the replacements (LOP `scatterinstances`, or the SOP `layoutbrush*` tools via `sopimport`) - staying silent would just move the hallucination one step later.

`test_version_annotations_are_actually_used` guards the syntax itself. A typo'd annotation degrades silently into an unconditional claim, which would then either pass for the wrong reason or fail on exactly the versions it was meant to exempt.

## Also

`test_full_usd_scene_assembly` built its stage output with `karmarendersettings`, failing the entire scenario on 20.5 rather than the single step that cannot work there. 20.5 has `karmarenderproperties`; `_first_available_lop()` picks whichever exists and the scenario asserts the same end state. It skips with a clear reason if neither is found, so a future rename fails loudly instead of quietly testing nothing.

The README claimed 20.5+ while only ever being exercised on 21.0 and 22.0 - which is how this went unnoticed. It now names the builds the suite actually passes on.

## Verification

Beyond the table above: unit suite 151 passed, ruff clean on both changed files. I also re-ran the plugin's own end-to-end checks on 20.5.278 and 20.5.654, which had never been exercised there - the #9/#15/loopback fixes and PR #14's USD `.keys()` and PDG-guard fixes all pass, confirming that USD change really is backward compatible against older USD.

One judgement call left open: the `(21.0+)` markers make the guidance *honest* on 20.5, not functional - a 20.5 user still cannot have Copernicus colour correction. If you would rather raise the supported floor to 21.0, most of these annotations could be deleted instead. That is a product decision, so I left 20.5 supported as the README claims.


healkeiser:
### Update: the annotations are now generated, and one Houdini install is enough

Two commits added since the original description. The hand-written markers in the first commit were the symptom; these remove the maintenance.

**`tools/gen_node_versions.py`** asks every installed Houdini for its own node type list, diffs them, and rewrites the `(X.Y+)` / `(X.Y-A.B)` markers in place. Humans still decide which names the instructions mention; the generator owns version truth. On its first run it found two names missed by hand (`apex::graph`, `apex::packcharacter`), for 34 annotations against 32.

Presence is authoritative, not SideFX's `#since`. The generator reports the comparison every run, and it is not close:

```
#since check : 4 corroborated, 2 differ, 28 undocumented
    Lop/karmarendersettings: derived 21.0, #since 18.0
    Lop/layout:              derived 20.5, #since 19.0
```

`#since` records when a node first appeared anywhere in Houdini's history and never records a removal, so building on it would have been wrong for `karmarendersettings` and blind to `instancer` and `layout` disappearing in 22. That comparison also corrected the wording: the note claimed `colorcorrect` "was added in 21.0", when all the diff proves is that it is absent from the oldest build sampled. A range now explicitly applies within the supported 20.5–22.0 window.

**The table accumulates, so a single install works.** The first version refused to run with fewer than two Houdinis, which made it useless to most contributors. `node_versions.json` now records which builds have been sampled and what each had:

| Scenario | Result |
|---|---|
| One install (21.0), table holds 20.5 + 22.0 | **All 34 annotations**, markdown identical |
| One install, empty table | Writes evidence, refuses to annotate, leaves markdown alone |
| Incremental merge vs one six-install pass | **Byte-identical** |
| Two full runs | Byte-identical (deterministic) |

`--check` changed meaning for the same reason: it now asks whether the committed table contradicts the builds installed here, which is answerable with one Houdini, rather than whether regeneration would produce identical files, which always looks stale to anyone with fewer builds.

**End users run nothing.** Verified by building the wheel: `server_instructions.md` ships with the annotations baked in; `tools/` and `node_versions.json` do not. The generator is a maintainer tool.

Worth stating plainly: the annotations were never the correctness mechanism. `build_network(dry_run=True)` already validates every node type against `category.nodeTypes()` of the running Houdini and returns did-you-mean corrections, which cannot go stale. The annotations reduce wrong first attempts, so a stale marker costs a round trip rather than a wrong answer.

### Platform coverage was thinner than it looked

Reviewing install discovery turned up three gaps, all fixed: `$HFS` was never consulted (the only way to find a relocated or studio-mounted install), Windows hardcoded `C:/Program Files` instead of `%PROGRAMFILES%`, and Linux only checked `/opt/hfs*`. Results are now de-duplicated too, since overlapping patterns and `$HFS` routinely name the same directory.

`tests/test_houdini_discovery.py` reproduces the Windows, macOS and Linux layouts under `tmp_path`, covering each platform's interpreter subpath, version ordering, de-duplication, `$HFS` and `%PROGRAMFILES%`. Only one layout can be exercised on any given machine, so this checks the logic everywhere - it does **not** prove SideFX's real macOS framework path is what it is believed to be, and that still wants one run on a Mac.

Two bugs surfaced while testing this. The `#since` merge was last-write-wins, so the committed file depended on the order contributors sampled in. And discovery split patterns on the `*`, walking one directory above the search root when a pattern's last segment was just `*` - harmless for the shipped patterns, and it would have bitten the next person adding one.

Also fixed three ruff violations of my own from the earlier commits here; I had only run a subset of rules on those files.

### Verification

All six installed builds still pass: 20.5.278, 20.5.487, 20.5.613, 20.5.654, 21.0.440, 22.0.368 - 123 passed, 1 skipped each. Unit suite 151 → 164. Ruff clean on everything touched.

One open suggestion, not built: ship `node_versions.json` in the wheel and have the server compare the connected Houdini's series against it at startup, so an unsampled version (a future 23.0) produces a visible "annotations may be stale" notice instead of silent drift.

PR #32, opened 2026-07-28
Give every shipped help corpus its own workflow guide

healkeiser:
## Why

The server instructs the assistant to read Houdini's own manual before improvising, then served **9 of the 47 help archives** the install ships. The 38 it skipped were exactly the workflow manuals that advice needs: `pyro/`, `fluid/`, `vellum/`, `destruction/`, `model/`, `assets/`, `copy/`. Copernicus, MPM, heightfields and ml were unreachable at *any* scope, because they ship as directories and the loader only read zips.

## What changed

**Help coverage.** `help_handlers.py` discovers scopes from the install and reads both layouts. 56 corpora, ~11,400 pages. Cost: 0.5s lazy load, ~69MB inside Houdini, and nothing in the assistant's context until a lookup happens.

**31 workflow guides, one per help scope**, replacing 6 generic prompts. Each is distilled from that scope's manual and cites the pages to read. `simulation_setup` dispatches per solver instead of covering pyro/FLIP/Vellum/RBD/MPM in one file; `houdini_workflow(topic)` serves the rest, so a new subject needs a markdown file and no code.

**No hand-written node names.** `tools/prompt_vocab.json` holds the editorial grouping; `gen_prompt_vocab.py` verifies all 382 names against `node_versions.json` and annotates version ranges. It caught real errors, several long-standing:

| finding | |
|---|---|
| LOP `payload`, `karmashadowcatcher` | not node types |
| `Dop/clothmaterial` | does not exist |
| `bulletsolver` | should be `bulletrbdsolver` |
| `Sop/twist`, `Dop/smokeobject` | deprecated in 22.0 |
| `ml_traindeformer` | a recipe, not a node |

**`gen_node_domains.py --check` could never pass.** It compared against its own un-annotated output while `gen_node_versions.py` adds markers in a second pass, so the "staleness" was the version markers themselves. Now compares ignoring markers, and warns that a plain run strips them.

**Layout.** `markdown/` splits into `instructions/`, `workflows/` and `shared/`, so a file's role is visible at every call site. Layout guidance moves out of Python constants into markdown.

**Quality.** New `Tests` workflow: lint, formatting, unit suite on Linux + Windows, generator checks. `ruff check` 212 findings to zero, including 24 missing exception chains added via an AST pass. Repo is `ruff format`-clean.

## Verification

- `ruff check` / `ruff format --check`: clean
- Unit suite: **361 passed**
- Integration suite in live Houdini 22.0.368: **125 passed, 1 skipped**
- `gen_prompt_vocab.py --check`, `gen_node_domains.py --check`, `gen_node_versions.py --check`: clean

The integration run caught one break the unit suite could not: `test_instruction_accuracy_live.py` still pointed at the pre-restructure path.

## Reviewer notes

- Two known compromises, both commented in `tests/test_prompt_vocab.py`: the prose name audit is a heuristic, so files using `camera`/`cache` as English carry a per-file collision list; and depth varies by corpus. Where SideFX documents judgment (pyro, fluid, vellum, mpm, destruction, copy, render, grains, props, character, muscles, composite) the guides carry verified craft. For `feathers`, `fur`, `io`, `anim`, `ml`, `heightfields_cop` I had page inventories more than page content, so those route to pages rather than asserting specifics.
- 12 help *directories* remain unreached only in the sense that `licenses`/`licensing`/`files`/`videos` are deliberately excluded as non-documentation.


healkeiser:
## ⚠️ This PR also fixes a live packaging break, found by the new CI

The new `Tests` workflow failed on all three matrix legs the moment it ran, with:

```
ModuleNotFoundError: No module named 'mcp.server.fastmcp'
```

Not a CI artefact. **mcp 2.0.0 has shipped and removed `mcp.server.fastmcp`**, which every tool, resource and prompt in this package imports. `pyproject.toml` declared `mcp>=1.8.0` with no upper bound, so:

**`pip install fxhoudinimcp` today resolves mcp 2.0.0 and the server cannot import at all.** That affects the currently published v2.6.2, not just this branch. My local environment has mcp 1.26.0, which is why nothing caught it until CI installed from a clean slate.

Fixed here by capping to `mcp>=1.8.0,<2`. Verified in a clean venv: resolves mcp 1.29.0, `import mcp.server.fastmcp` succeeds.

Lifting the cap later is a real port, not a range widening: it means moving to whatever replaced FastMCP in 2.x. Worth its own issue.

## Status

| check | |
|---|---|
| Lint | pass |
| Test (ubuntu, py3.10) | pass |
| Test (ubuntu, py3.13) | pass |
| Test (windows, py3.13) | pass |
| Generated content up to date | pass |

Also rebased onto current `main` (the branch was cut from a 9-commit-stale base). The rename-vs-edit conflict on `server_instructions.md` was resolved by hand so the throughput passage from #31 survives inside the compressed rule 1 — both the 50 ms/800 ms-vs-66 ms argument and the new help-corpus guidance are present. Integration suite re-run on the merged result: 125 passed, 1 skipped, in live Houdini 22.0.368.

PR #35, opened 2026-07-29
Stop tools reporting success they never checked, and test that they admit failure

healkeiser:
**Draft until validated against a graphical Houdini.** See "What is not verified" below — shipping a false-success fix without confirming it against a real viewport would repeat the mistake it exists to correct.

## Where this came from

A second session in `fxhoudinimcp-test` built an ocean in Solaris and ended with the user listing what was wrong: *"Depth is inexistant, I barely see any waves, camera framing is completely different, I haven't seen any realistic shading."* `execute_python` went from **13 calls to 32**.

The cause was mostly not missing capability. It was **tools that lied.**

## 1. `set_viewport_renderer` reported success it never checked

It returned `{"success": true}` whenever a setter didn't raise, and never read the renderer back. So it reported Karma while the viewport sat on GL — and every screenshot the assistant verified against was the wrong image. Nine `execute_python` calls went into discovering that; the model's own words were *"only sets it and previously reported success without effect"*.

Now uses the documented pair `hydraRenderers` / `setHydraRenderer` / `currentHydraRenderer`, reads the active delegate back, and **raises** when the viewport is not on the requested renderer. Where a build has no readback it returns `verified: false` with a note instead of claiming the request took.

## 2. `set_viewport_camera` could not frame a Solaris shot at all

It resolved through `hou.node()` alone. A USD camera prim is not a node, so it raised "Camera node not found" for a perfectly valid stage path. Now accepts either, and confirms with `cameraPath()` — because `setCamera` can return quietly and leave the viewport on free perspective, which is exactly what happened.

`get_viewport_info` now reports `camera_path` and the active `renderer`, so "what am I looking through, and with what" is one call. Neither was reported before.

## 3. `start_render` refused everything except `/out`

It rejected any node whose category wasn't `Driver`, which ruled out:

- the **LOP `usdrender_rop`** — how Solaris renders
- SOP ROPs (`rop_geometry`)
- a **File Cache's Save to Disk**

Ten `execute_python` calls pressed those by hand. The test is now whether a node *can be executed*: `render()` where it exists, otherwise its `execute` button with the frame range applied to `trange`/`f1..f3`. It also reports the output path and **whether anything is on disk there** — which the session resorted to PowerShell to check.

## Verified headless

- 423 unit tests
- 149 integration passed, 1 skipped, **183/183 command coverage**
- 5 new tests: execute a SOP ROP and confirm the file, press a File Cache button, accept a LOP `usdrender_rop`, and give a useful error for a node with nothing to trigger (asserting the old "not a ROP/Driver node" wording is gone)

## What is NOT verified

The viewport changes cannot run in hython. `tests/integration/gui_session_check.py` gained checks that:

- **FAIL** if `set_viewport_renderer` claims a delegate the viewport does not report
- **FAIL** if `set_viewport_camera` accepts a bad path instead of raising
- report `renderer` and `camera_path` from `get_viewport_info`

I could not confirm the exact API route for binding a **USD camera prim** to a Solaris viewport. `GeometryViewport.setCamera` documents `(camera_node, camera_name)`; the string-prim overload is attempted and then *verified*, so a route that does not work now fails loudly instead of silently. That is the intended behaviour either way, but whether Solaris framing actually succeeds needs a GUI run to confirm.


healkeiser:
## Validated against a graphical Houdini 22.0.368 — and it found two defects

Which is the entire reason for gating a false-confidence fix on a real viewport.

### 1. My verification was circular

`cameraPath()` is a **pure echo**. Confirmed live:

```
setCamera("/cameras/does_not_exist")  ->  no error
cameraPath()                          ->  '/cameras/does_not_exist'
camera()                              ->  None
```

So comparing the echo to the requested path always passed. The first commit **accepted a bogus prim and reported `verified: true`** — the same class of bug it was written to fix.

Verification is now a **stage lookup**: the prim must exist and be a `Camera`. The OBJ path verifies against `viewport.camera()`, which returns the node rather than a string.

### 2. "No readback" was the wrong diagnosis

`currentHydraRenderer()` raises `TypeError: Specified view is not a scene graph view` in an object-level viewport, because Hydra delegates are a Solaris concept. My note said *"This Houdini exposes no currentHydraRenderer()"*, which would send someone looking for a Houdini upgrade instead of a LOP network. The two cases are now distinguished.

In Solaris the readback works fine: `hydraRenderers()` → `('Houdini VK', 'Karma CPU', 'Karma XPU', 'Storm')`.

### Live results

| check | result |
|---|---|
| renderer verified | `Houdini VK` → `Karma CPU` → `Storm`, read back each time |
| renderer in `/obj` | raises, naming the reason |
| USD camera prim | binds and verifies, `prim_type: Camera` |
| missing prim | raises: *No prim at '/cameras/does_not_exist'. Camera prims present: ['/cameras/mcp_gui_lopcam']* |
| non-camera prim | raises: *'/mcp_gui_geo' is a Sphere prim, not a Camera* |
| `get_viewport_info` | reports `renderer`, `camera_path`, `camera_path_resolves` |

**GUI gate: 18 passed, 0 failed.** The gate now builds and removes its own Solaris stage, so the next change to these tools has something to fail against.

Headless unchanged: 423 unit tests, 149 integration, 183/183 command coverage.

Also: `get_viewport_info` now reports `camera_path_resolves`, because it was repeating the same unchecked echo and would have become a second source of the same false confidence.

Ready for review.

healkeiser:
## Items 4–6 added. 183 → 188 tools, new Shelf Tools category

### Shelf tools (7 of the 32 escapes)

Some setups are not built from nodes. `list_shelf_tools` finds them among ~8,000 installed, `get_shelf_tool_script` shows SideFX's own recipe and **names the toolutils module holding it**, `run_shelf_tool` runs it and reports what it created.

Verified live: `run_shelf_tool("geometry_largeocean")` created `/obj/ocean_interior` and `/obj/ocean_surface` **in one call** — the thing the ocean session spent seven `execute_python` calls reimplementing. Headless it fails with a sentence naming `hou.ui` and pointing at the script reader, instead of an `AttributeError` that reads like a server bug.

### `set_viewer_context` (3 more)

`set_current_network` moves the network **editor**; this moves the **viewer**, which is what decides whether a scene graph view exists. No scene graph view means no Hydra delegate and no bindable USD camera, so this is the first call in any Solaris preview. The GUI gate now uses it instead of `execute_python`, so it dogfoods and covers itself.

### `get_cook_status`

Answers the after-the-fact question: did this recook, is it time dependent, did it error, unsaved changes. It deliberately **does not pretend to be a progress poll** — every command runs on Houdini's main thread, so a long cook blocks the bridge, which is exactly why the session watched process CPU from PowerShell. No tool can change that shape, and claiming otherwise would be the false confidence this branch exists to remove.

### Expression tools: a docs fix, not a code fix

The session claimed no tool sets expressions, then destroyed a File Cache's derived output path with a literal sweep. **`set_expression`, `get_expression`, `link_parameters` and `revert_parameter` all already existed** — it just did not find them among 183 tools. They are now named in `TOOL PRIORITY`.

### And one removal, which is the interesting part

I built `find_broken_expressions` for that same gap. Then I tested it against a **real** network instead of a synthetic one:

| version | result on a healthy stock ocean setup |
|---|---|
| pattern-matched self-references | **62 "broken" parameters** — all false positives |
| narrowed to expressions that raise | 0 false positives, and **0 true positives** |

Houdini surfaces a bad expression at **cook** time, not on read. And `verify_network` already reports exactly that, naming the parameter:

```
Unable to evaluate expression (Unknown function in expression (/obj/geo1/box1/sizex))
```

So the capability existed under another name. The tool is deleted and the instructions point at the cook. A tool that sends a caller on a repair spree across a working network is worse than no tool — that spree is what broke the File Cache in the first place.

### Verification

- 423 unit tests
- 159 integration passed, **188/188 command coverage**
- GUI gate: **18 passed, 0 failed** against Houdini 22.0.368
- 8/8 CI checks green

healkeiser:
## Real render attempted. It found two more bugs, and one claim I cannot make.

Following up on "do a real render to confirm, do not leave lingering possible bugs".

### The render could not be completed here

I built a full Solaris stage in the live session (sphere -> domelight -> camera -> `karmarendersettings` -> `usdrender_rop`), dropped it to 160px with 1 sample per pixel so it would finish in seconds, and pointed `outputimage` at a temp path. `start_render` returned in 0.5 seconds. Nothing was written.

The ROP's error said why:

```
Command Exit Code: 3
No licenses could be found to run this application.
```

`usdrender_rop` shells out to `husk`, and husk is not licensed on this machine. **A completed Karma frame on disk is the one thing in this PR I have not verified, and cannot verify here.** It needs a licensed machine.

### But the attempt was worth more than the frame would have been

`start_render` reported `success: True` for that failed render. This is the same false-confidence bug this PR exists to fix, sitting inside the tool I added to fix it. `node.render()` returns normally when the ROP it delegates to fails; the failure only lives in `node.errors()`, which nothing read. A caller would have gone on to screenshot an image that was never written, and rule 3 would have been unsatisfiable again.

Three fixes:

- **Read `node.errors()` after rendering.** `success` now reflects them, and the errors are reported.
- **Snapshot output paths before and after.** `wrote_files` answers "did anything appear or change", not "does a file exist" -- otherwise re-rendering over yesterday's frame reads as success. The raising path returns the same shape, so no caller branches to find out whether output landed.
- **Stop calling `__render__.usd` and MPlay's `ip` missing files.** They are not files. Reporting them as `exists: False` teaches the reader to ignore the field, which is how the first bug goes unnoticed.

### Second bug: `run_shelf_tool` watched one network

`largeOcean` builds in four. Verified live: a procedural in `/stage`, two material builders in `/mat`, a fetch ROP in `/out`. It reported 2 of them, and **six nodes had silently accumulated** across earlier runs that no caller was ever told about. It now watches `/obj`, `/stage`, `/out`, `/mat`, `/img` and reports the whole set. All five networks are back to 0 children.

### Withdrawing an earlier claim

I said in this branch that `get_scene_info` reports empty `*_children`. That was wrong -- it has no such keys, and the empty values came from my own cleanup script. No bug there.

### Coverage of the failure branch, without a license

The bug above existed because no test ever made a render fail. Now, all headless:

- a ROP pointed at an impossible drive must report `success: False`, carry the error, and `wrote_files: False`
- a working ROP must report `wrote_files: True` and a file whose mtime actually changed
- a non-file target must be `is_file_output: False`, not missing

One of those tests wrote a file named `ip` into the repository root on its first run, because a SOP ROP treats `"ip"` as a relative path rather than an MPlay target. It briefly reached a commit; it is gone from the branch history, the test no longer executes anything, and the path is gitignored.

### Gates

ruff clean, 126 files formatted, 423 unit tests, 162 integration passed, 188/188 command coverage, all nine CI checks green (mcp 1.14.0 / 1.29.0 / 2.0.0).

healkeiser:
## Fixed all 3, and found 3 more while doing it

### The two you asked about

**`export_file`'s Driver branch** returned the hardcoded string `"Render complete."` after `node.render()` -- no errors read, no file checked. It now reports `success` / `errors` / `wrote_files` like `start_render`.

**`write_cache`** set `status = "success"` after `pressButton()`. This one was worse than it looked. Verified on 22.0:

```
filecache errors      : ()
  CHILD /obj/probe/filecache1/render (rop_geometry) errors:
    ('Failed to save output to file "Q:/.../c.0001.bgeo.sc".', ... x10)
```

A filecache is an HDA that presses an **embedded ROP**. It reports `errors() == ()` on itself while the node doing the work holds every error. So reading the top node -- which is what my first fix did -- would still have missed the entire failure. `node_messages` now walks inside when the node is silent (bounded: 400 descendants, 12 messages) and prefixes each error with the node it came from, because `/render` inside a filecache is not somewhere a caller thinks to look.

**The frame restore** in the SOP loop: done, and asserted.

### Three more, found while fixing those

- **`export_file` ignored `file_path` entirely for ROPs.** It rendered to whatever the node's output parm already said, while the docstring promised `file_path` was the destination. It now points the ROP at what was asked for, and restores the original path afterwards -- repointing is a means, not the request.
- **Frame naming broke on compound extensions.** `os.path.splitext("seq.bgeo.sc")` returns `("seq.bgeo", ".sc")`, so frames were written as `seq.bgeo.0001.sc` -- a name Houdini writes but no sequence reader recognises. Now `seq.0001.bgeo.sc`.
- **Silence was still being treated as success.** My own fix from yesterday had this: no errors + no file = `success: True` with a note. A node told to write a file that reports no error and produces nothing did not do its job. `success` now requires evidence, and where no file target is recognised at all it says `output_verified: False` rather than implying a check happened.

### On your convention question

You asked whether readback should be the convention or per-tool judgment. I went with the convention, because per-tool judgment is what produced three different answers to the same question. It lives in `fxhoudinimcp_server/outputs.py` -- one `write_verdict`, one `failure_verdict`, one place to fix the next thing we learn. Existing shapes are preserved: `write_cache` still reports `status`, now derived from the same evidence as `success`, and a test asserts the two can never disagree.

### Why 188/188 didn't catch any of this

`write_cache` was already covered -- by `call("cache.write_cache", ..., allow_error=True)`, which asserts nothing about the answer. Coverage there means *invoked*, not *made to fail*. The eight new tests all assert what the report **says**: an unwritable drive must come back `success: False` with the error and `wrote_files: False`; a real write must show a file that actually changed; the ROP's output path must be restored; a frame-range export must leave the playbar where it found it.

### Deliberately not fixed

`import_file`'s `buildHierarchy` `pressButton()` is the last instance of the pattern. I inspected it and left it: its `success` refers to a node that demonstrably exists, and there is no `.abc` fixture here to verify a change against. Guessing at it would repeat the mistake this branch is about.

423 unit, 170 integration, 188/188 coverage, nine CI checks green.

Issue #2, opened 2026-05-12
Auto Layout should be a toggle


rwoodsfx-ctrl:
This combined with the fact throughout every single script layout is used over and over is awful for people who actually use their scenes. 

> Call layout_children frequently — after every batch of 3-5 new nodes, not just at the end. A tidy graph lets the user follow along in real time.

Please consider implementing a toggle feature. While the MCP has demonstrated promising results thus far, its potential to disrupt my networks and ruin intentional layouts for various projects is a cause for concern. Consequently, I am hesitant to deploy it in any practical applications.

PR #37, opened 2026-08-13
Place new nodes instead of stacking them at the origin

d-potrivaev:
### The problem

#2 asked for a toggle so the MCP would stop rearranging intentional layouts, and `FXHOUDINIMCP_AUTO_LAYOUT` delivered one. But turning it off swaps one layout problem for another: with the flag off, nothing positions a new node at all.

- `create_node` sets a position only when the caller passes one.
- `build_network(layout=True)` — the default — calls `layout_if_enabled`, which the flag turns off. So the parameter promises a layout and silently does nothing.

Everything a session builds therefore lands at `(0, 0)`, one node on top of the next. Measured on Houdini 22.0.368 with the flag off, on scenes saved by the agent itself: a 46-node container and a 31-node container, **100% of nodes at the origin in both**.

There is a second, smaller defect that bites with the flag either way. The handlers that *do* position a node call `moveToGoodPosition()` bare — and its `move_inputs`, `move_outputs` and `move_unconnected` all default to `True`, so placing one new node can drag the user's existing nodes around. That is what #2 objected to in the first place.

### The fix

One helper, `config.place_new_node()`, places the node that was just created with those three kwargs pinned to `False`. It is deliberately **not** gated by the flag: the flag guards *rearranging an existing network*, not giving a brand new node somewhere to stand.

- `create_node` uses it when no position was given; an explicit position still wins.
- `build_network(layout=True)` places the nodes that call created, in creation order, so a fresh chain reads top to bottom. Nodes that already existed are never touched, which makes building into a hand-arranged network safe. With the flag on, the parent is still laid out afterwards exactly as before.
- The nine bare `moveToGoodPosition()` calls in the CHOP, LOP, material and rendering handlers go through the helper, and a source-level test keeps a bare one from coming back.

### Verification

On Houdini 22.0.368, against a container whose nodes were arranged by hand: a seven-node addition built through `build_network` and `create_node` left every hand-placed node exactly where it was, put nothing at the origin, and produced no two nodes sharing a position. A node with no inputs (a CTRL null) takes a free column rather than the origin.

`layoutChildren(items=[...])` looked like the obvious tool for "lay out only the new nodes" and turned out to be the wrong one: measured on the same build, it lays the subset out as a standalone graph anchored at the origin and drops it on top of the nodes it was told not to touch. Hence per-node placement rather than a subset layout.

Tests: 8 new, covering the pinned kwargs, creation order, an explicit position winning, placement surviving a `moveToGoodPosition` failure, and the source guard. Full suite green on this branch (431 passed).

### Not in this PR

The default value of `FXHOUDINIMCP_AUTO_LAYOUT` is untouched — that is a policy call, and this fix works either way.


healkeiser:
Thanks, the diagnosis here is right and the measurements are useful. `layout=True` funnelling into a flag-gated no-op is a real bug, and pinning the three `move_*` kwargs is the correct read of #2. Three things before I merge.

**1. Placement shouldn't be gated by `layout`.** `FXHOUDINIMCP_AUTO_LAYOUT` governs auto-layout only, which affects hand-created nodes as much as agent-created ones. Placement is a floor, not a layout option: no matter the flag's value, and no matter `layout`, a newly created node must not stack at the origin. So `place_new_nodes` should run unconditionally in `build_network`, with `layout` continuing to gate only `layout_if_enabled(parent)`.

**2. It needs to cover every creation path, not two.** `workflow_handlers.py` alone has 61 `createNode` calls (`setup_pyro_sim`, `setup_flip_sim`, `setup_rbd_sim`, `setup_vellum_sim`, `setup_render`, `build_sop_chain`), and the `cop`, `vex`, `scene` and `geometry` handlers are in the same position: all of them still rely on `layout_if_enabled` alone, so with the flag off they keep piling at the origin. Note that `TestHandlerSourceGuard` passes trivially for those files, because they never called `moveToGoodPosition` in the first place, so the guard currently reads as coverage it doesn't have.

The chokepoint worth using: every one of those paths already ends in `layout_if_enabled(parent)`, either directly or through `_focus_network_editor` (`cop_handlers.py:57`, `scene_handlers.py:34`, `workflow_handlers.py:37`, same in the rest). Putting the placement there fixes all eleven modules at once, and covers whatever handler gets added next. Placing after wiring is also what `relative_to_inputs` needs, so the existing creation-order sequencing still applies.

**3. CI lint is red.** `ruff` wants `contextlib.suppress(Exception)` over `try`/`except`/`pass` at `config.py:60` (SIM105). One line, and it matches the idiom already used in `graph_handlers.py`.

`moveToGoodPosition` with those kwargs is the right tool and I am not asking you to replace it. One thing to confirm when you re-measure, since Houdini is not available on my side right now: new nodes must land below what they follow, never above. A wired node under its inputs is what you already measured. The case I would want a second look at is the input-less node taking a free column, which is fine as the head of its own chain but should not end up level with or above existing nodes.

healkeiser (changes requested):
Requesting changes on the three points in https://github.com/healkeiser/fxhoudinimcp/pull/37#issuecomment-5362196086: placement should not be gated by `layout`, it needs to cover every creation path via the `layout_if_enabled` chokepoint, and CI lint (SIM105) is red. Diagnosis and the helper itself are good, so this stays open as the base rather than being closed.

d-potrivaev:
All three points addressed, and the re-measure turned up a fourth hole worth closing.

**1. Placement ungated from `layout`.** `build_network` now places what it created unconditionally; `layout` keeps gating only `layout_if_enabled(parent)`. The tool docstring says so.

**2. The chokepoint.** `layout_if_enabled` is now the floor: with auto-layout off it places any child still parked at (0, 0) — Houdini's mark of a node nothing has positioned — in `children()` (creation) order, after wiring, since the chokepoint runs at the end of each handler; with the flag on, `layoutChildren` covers them as before. That picks up all eleven modules and whatever handler comes next. One cost worth naming: a node someone deliberately parked at exactly (0, 0) gets nudged once. One path never reached the chokepoint at all — `set_detail_attrib` in geometry_handlers wires an attribcreate and positioned nothing — it now places explicitly. You were right about the guard: it got a companion with teeth (any handler source containing `createNode(` must reference `place_new_node`, `layout_if_enabled` or `_focus_network_editor`), plus behaviour tests on the chokepoint itself.

**3. SIM105** — `contextlib.suppress(Exception)`; `ruff check` and `ruff format --check` both pass.

**The fourth hole.** The workflow handlers call `layout_if_enabled(geo)` on a container nothing ever positioned, and `layoutChildren` never moves the parent — so the pyro chain landed fine *inside* the container while the container itself sat at (0, 0) under `/obj`, and a second `setup_*_sim` stacked its container on top (measured). `layout_if_enabled` now also places the node it was handed when that node is still at the origin, in both flag states — guarded for top-level managers, because `moveToGoodPosition` on `/obj` happily relocates it inside the root network (measured on 22.0.368).

Re-measured on 22.0.368 with auto-layout off, against a hand-arranged network (rows at y = 3.0 / 1.5 / 0.5):
- your question first: an input-less node lands **below** everything, in a free column — (1.69, −0.39), then a second at (3.39, −1.29); never level with or above existing nodes. The sole node of an empty network stays at the origin as the head of its own chain.
- hand-placed nodes kept their exact coordinates through `create_node`, `build_network(layout=False)` and two `setup_pyro_sim` calls;
- `build_network(layout=False)` still places its merge/color/null chain below its inputs, nothing at the origin;
- a third pyro container lands at (5.39, −2.79) under `/obj` instead of joining the pile, and `/obj` itself does not move;
- flag on: `layoutChildren` runs exactly as before.

Full suite green locally (440 tests).

PR #42, opened 2026-09-12
Close the holes the placement floor left open

healkeiser:
Follow-up to #37, which was merged with the review findings still open. Six things, in the order they bite a user.

### The floor fired on handlers that create nothing

`connect_nodes`, `connect_nodes_batch` and `set_node_flags` all route through `_focus_network_editor`, which reaches `layout_if_enabled`. With `FXHOUDINIMCP_AUTO_LAYOUT=0`, a call that only rewired two nodes relocated whatever the user had parked at (0, 0). That flag's whole purpose is preserving hand layouts, so those three now pass `place_unpositioned=False` and the floor stays where it belongs: freshly created nodes.

### Position alone could not say whether a node had been placed

`moveToGoodPosition` legitimately leaves the first node of a network where it is, so "sits at (0, 0)" marked that node unplaced for ever. A pyro container placed at the origin jumped later, on an unrelated call that only touched its insides. Placement now records itself in user data, which the `.hip` carries across a reload. Same tag makes an explicit `position=[0, 0]` stick, where the floor used to read it back as "never positioned" and move the node the caller had just placed.

### Two networks never reached the floor at all

- `_setup_rbd_sim`'s legacy DOP branch creates five nodes inside `dopnet1` and only called `layout_if_enabled(geo)`. `layoutChildren` does not descend, and the origin sweep only walks direct children, so those five stayed stacked with the flag in either state.
- `setup_render` creates its camera in `/obj` and only laid out `/out`, so back-to-back calls piled `render_cam`, `render_cam1`, ... at the origin.

### The guard test now guards

The previous check asked whether `place_new_node` / `layout_if_enabled` / `_focus_network_editor` appeared anywhere in a handler file. Both bugs above live in `workflow_handlers.py`, which mentions all three elsewhere, so it passed vacuously, the same failure noted in the #37 review one level up. It is now per network created into: every receiver of a `createNode` call must be handed to the floor, with nine documented exemptions (a scratch probe, the root-level `/mat` context, delegating helpers, and containers whose sole child placement would not move anyway). A second guard keeps the floor off handlers that create nothing.

Both guards were confirmed to fail with each bug reintroduced, and each new behaviour test fails with its own fix reverted.

### Docstring

`build_network`'s `layout` argument no longer claims to gate placement; placement runs unconditionally, and the client-side docstring already said so.

---

445 tests pass. `ruff check` and `ruff format --check` are clean.

Verified on hython 22.0.368, headless, with the flag off: 9/10 checks pass, and the tenth is a separate pre-existing bug in the RBD fallback (dopimport::2.0 takes no input). Numbers in the comment below.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


healkeiser:
Verified against a real Houdini after all: hython 22.0.368, headless, `FXHOUDINIMCP_AUTO_LAYOUT=0`, exercising this branch's handlers directly. 9/10 checks pass, and the tenth is explained below.

**Baseline still holds**
- a wired node lands under its input at (5.0, 4.1058) and the hand-placed box at (5.0, 5.0) does not move;
- an input-less node takes a free column at (6.6944, 3.2116), below the network, never level with or above it.

**The floor no longer fires on handlers that create nothing**
- `connect_nodes`: a null parked at (0, 0) stays at (0, 0);
- `set_node_flags`: same;
- positive control, so this is gating and not a dead floor: calling `_focus_network_editor` without the new flag moves that same parked node to (6.6944, 2.1058).

**Placement records itself**
- a container placed at the origin (where `moveToGoodPosition` correctly leaves the first node of a network) stays at (0, 0) through a second container going to (1.6944, -0.8942) and through a later call that only touched its insides. Before the tag this was the jump reported in review;
- `userData("fxhoudinimcp_placed") == "1"` on the placed node.

**`position=[0, 0]` survives the floor** - the node stays at (0, 0) next to an existing node at (2, 2).

**`setup_render`** - two calls give `render_cam` at (0, 0) and `render_cam1` at (1.6944, -0.8942), instead of both at the origin.

**RBD legacy DOP fallback** - the SOP Bullet solver exists on 22.0, so I fault-injected its `createNode` to force the fallback. All six dopnet children are placed and tagged, none left at the origin:

```
file_output   (1.6944, -0.8942)    rbdobject1  (3.3888, -1.7884)
rbdsolver1    (0.0, -0.8942)       merge1      (1.6944, -2.6826)
groundplane1  (5.0832, -2.6826)    gravity1    (0.0, -1.7884)
```

### Separate pre-existing bug, found while forcing that path

The fallback then raises `hou.InvalidInput` at `workflow_handlers.py:564`:

```python
dopimport = geo.createNode("dopimport", "dop_import1")
dopimport.setInput(0, last_sop, 0)   # InvalidInput
```

`createNode("dopimport")` resolves to `dopimport::2.0` on Houdini 22, and that type has `maxNumInputs == 0`; version 1 had one input. So `_setup_rbd_sim`'s legacy branch cannot complete on 22.0 at all, leaving a half-built network behind. Unrelated to placement, untouched by this PR, and worth its own issue.

healkeiser:
Fixed the `dopimport` bug here too, since it sits in the branch this PR already touches.

`createNode("dopimport")` resolves to `dopimport::2.0`, whose `maxNumInputs` is 0. It reads the sim out of its `doppath` parm; only the version 1 node took an input. So `dopimport.setInput(0, last_sop, 0)` could never work on a modern build, and `_setup_rbd_sim`'s legacy branch died half-built on every Houdini 22 run. The pyro (`:307`) and flip (`:710`) paths in the same file already treat the DOP Import as the head of its own chain, so the fix is to match them.

Measured on 22.0.368 with the legacy branch forced:

```
PASS  the legacy DOP fallback completes
PASS  DOP Import points at the dopnet   doppath='/obj/rbd_sim/dopnet1'
PASS  dopnet children placed (6 children, 0 at origin)
PASS  cook at frame 1: no errors

SOP chain in /obj/rbd_sim:
  source_geo    object_merge              inputs=[]
  fracture1     rbdmaterialfracture::4.0  inputs=['source_geo']
  dopnet1       dopnet                    inputs=[]
  dop_import1   dopimport::2.0            inputs=[]
  file_cache1   filecache::2.0            inputs=['dop_import1']
```

The regression guard is AST-based rather than a text match: reading *from* a DOP Import is legitimate (`filecache.setInput(0, dopimport, 0)` appears twice and is correct), so it flags only a `setInput` whose receiver is a variable assigned from `createNode("dopimport"...)`. Confirmed to fail with the line reintroduced and to pass otherwise.

446 tests, `ruff check` and `ruff format --check` clean.

PR #68, opened 2026-09-14
Say what the installer actually writes: the plugin path, no settings

healkeiser:
Found while rerunning `fxhoudinimcp install` after #67. The README's "Configuring the plugin" section said the package file the installer writes ships every Houdini-side setting at its default and is the place to change them. `package_json()` has only ever written the `FXHOUDINIMCP` path entry; the settings block exists in `houdini/fxhoudinimcp.json` in the repo and nowhere else. So the v2.16.0 note about a stale `"1"` in users' package files was wrong too, and has been corrected on the release.

- README: the settings are environment variables with defaults, set in `houdini.env` or as an entry added to the package file's `env` list; the repo template shows the shape. Precedence warning kept.
- configuration.md: two mentions of "the package file the installer writes" made accurate.

Docs only. If the README's original design is what you want, the installer writing every setting explicitly, that is a separate change to `package_json()` and I would rather it be a decision than a side effect.


healkeiser:
Superseded: the installer will write the settings block and keep customised values, so the docs will describe that instead.

PR #70, opened 2026-09-15
link_parameters writes chs() for a String destination and a path relative to the destination node

d-potrivaev:
`link_parameters` builds the reference as `ch("<absolute source path>")` whatever the destination is. Two things go wrong with that.

`ch()` evaluates the referenced channel as a number, so a String destination reads back `"0"`: four String spare parameters inside an HDA (material paths on a control null) were all `"0"` after linking, and the session fell back to `set_expression` by hand for each. And an absolute path breaks as soon as the pair is moved, collapsed into a subnet or saved into an HDA and instanced elsewhere. Houdini's own Paste Relative References writes `chs("../CONTROL/mat")`.

### Change

The channel function follows the destination's template type: `chs()` for `String`, `ch()` for floats, ints, toggles and int-valued menus. The path is `dst.node().relativePathTo(src.node())` plus the parameter name; a link within one node is the bare name (`ch("tx")`). The reply reads the destination back so the caller does not need a second call: `function`, `relative`, `value`, and a `warning` when a String destination is linked to a numeric source (`chs()` converts on read).

### Verified on 22.0.429

`null CONTROL` with a String spare `mat = /obj/shopnet1/brick` → `null n2` String spare `label`: expression `chs("../CONTROL/mat")`, `get_parameter(n2, label)` → `"/obj/shopnet1/brick"` (was `"0"`). Int → Int: `ch("../CONTROL/count")` = 3. Same node: `ch("tx")`. After `collapseIntoSubnet` of both nodes the linked values are intact, which an absolute path would not survive.

### Tests

`tests/test_link_parameters.py`: String destination gets `chs()` and a relative path, float keeps `ch()`, same-node link is a bare name, a mixed pair is named in a warning. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: 6a6c856. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #71, opened 2026-09-15
get_node_info compares a tuple component with its own default, not the whole tuple

d-potrivaev:
`get_node_info` reports `non_default_parameters` to keep the reply compact, but the comparison was wrong for every tuple: `parm.eval()` (a scalar) against `parmTemplate().defaultValue()` (the whole tuple). `ty` at 0.0 compared with `(0.0, 0.0, 0.0)` is "not equal", so every component of every tuple came back as an edit.

### Reproduction

A bare `xform` with `tx = 1`: **34** non-default parameters, `ty`, `tz`, `rx…`, `sx…`, both pivots, `prexform` and two buttons, each with `value: 0.0, default: [0.0, 0.0, 0.0]`. A `geo` object with nothing touched: 23. A MaterialX node: about 12 KB of untouched parameters.

### Change

The summary loop moves into `_non_default_parms`. Houdini's own answer, `parm.isAtDefault()`, is asked first (it also sees an expression as a change); the fallback compares the component with `defaultValue()[componentIndex()]`. Buttons, folders, folder sets, separators and labels carry no value and are skipped. The Ramp/Data behaviour is unchanged: a value that cannot be compared still reaches the reply, which is why it has to survive the JSON encoder.

### Verified on 22.0.429

`xform` with `tx = 1` → one entry, `["tx", 1.0, 0.0]`, out of 49 parameters. Untouched `xform` and `geo` (90 parameters) → none.

### Tests

`tests/test_node_info_defaults.py`: one edited component is the only entry and its default is the component's, Houdini's `isAtDefault` wins when it answers, a Ramp without a comparable default still reaches the reply. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: 33b6aaa. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #72, opened 2026-09-15
USD arrays come back summarised unless asked for in full; get_usd_attribute reads windows

d-potrivaev:
`get_usd_prim` and `get_usd_attribute` return every array in full. On a building's mesh prim the reply was **6,598,966 characters** (`points`, `primvars:st`, `faceVertexIndices`); `get_usd_attribute(primvars:st)` alone was 643 KB over 40,000 lines. Nobody reads that; the session saved it to a file and grepped it. What a mesh question needs is how many, of what, the first few and the range.

### Change

`_usd_value_to_python` takes `array_limit` (default 16). An array longer than that comes back as `{array: true, size, element_type, head: [first 8], min, max, note}`; `min`/`max` are per component for vector arrays (numpy, which Houdini ships). Short arrays such as `extent` and `xformOpOrder` are untouched.

`get_usd_prim(full=False)`: `full=True` returns every element, and the reply says `arrays_summarised`. `get_usd_attribute(full=False, offset=0, limit=64)`: a long array answers with the summary as `value` plus `slice: {offset, limit, count, has_more, values}`, so a big array is read in windows; `full=True` returns the whole array as `value`. The client tool docstrings say so.

Found on the way: with pxr importable, a token attribute (`orientation`, `purpose`, `visibility`) went through the generic Vt fallback character by character and came back as a nested list, about 24 KB per token on a Mesh; a `str` now returns before any array handling.

### Verified on 22.0.429

`sopimport` of a 100×100 grid, `get_usd_prim(/grid_import/mesh_0)`: **2,976 characters** (`points.size 10000`, `min [-5, 0, -5]`, `max [5, 0, 5]`, `faceVertexIndices.size 39204`) against **729,427** with `full=True`, where `points` is the full 10,000. `get_usd_attribute(faceVertexIndices)`: 705 characters, `slice.count 64`, `has_more: true`; `points` at `offset=9990` → 10 elements, `has_more: false`; `extent` whole.

### Tests

`tests/test_usd_arrays.py`: a short array passes whole, a long one is a summary with the right head and size, `full` returns everything, text is never an array (also with pxr present), the default window is the first 64, offset/limit walk to the end, `full` drops the slice, a short array has none. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: 62ec2a1. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #73, opened 2026-09-15
Cooked-evidence replies say when the update mode keeps nodes from cooking

d-potrivaev:
With the update mode set to Manual, nothing cooks on its own, so a verb that reads cooked evidence answers with zeros and `success: true`: `get_geometry_info` reports 0 points on a fresh node, `build_network`'s `geometry` block says `0/0`, and nothing in any reply says why. The mode was not visible either; `get_scene_info` did not carry it, and the only way to learn it was `set_update_mode()` with no argument.

Cooking the node from the handler is not an option: measured on 22.0.429, in Manual mode neither `node.cook(force=True)` nor `opcook -F` nor `geometryAtFrame()` produces geometry (`needsToCook()` stays true). The honest contract is to say so.

### Change

`config.update_mode_name()` returns `auto` / `on_mouse_up` / `manual` (the tokens `set_update_mode` takes), `config.update_mode_warning()` phrases one sentence when the mode is not Auto Update. `get_geometry_info` attaches it as `warnings`, `build_network`'s geometry summary as `warning`, and `get_scene_info` gains `update_mode`.

### Verified on 22.0.429

Manual: `get_scene_info.update_mode = "manual"`; `get_geometry_info` on a fresh box → `num_points 0` with the warning; `build_network` of a sphere → `geometry.warning` set. Back to Auto: `update_mode = "auto"`, `num_points 8`, no warning.

### Tests

`tests/test_update_mode_warning.py`: Auto gives no warning, Manual is named and points at `set_update_mode('auto')`, an unanswerable `hou` is not an error, the geometry summary carries the warning only when there is one. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: 54496a6. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #74, opened 2026-09-15
list_usd_prims and get_usd_prim_stats treat the root as a path, not a string prefix

d-potrivaev:
`list_usd_prims(root_path=...)` and `get_usd_prim_stats(prim_path=...)` decide "is under the root" with `str.startswith`. `/materials/BLD` therefore matches `/materials/BLD_probes` as well: two `primitive` LOPs with those paths answered `count: 2` for a root that holds one.

### Change

Both walk `Usd.PrimRange(root)` (the stage traversal stays for `/`) and test membership with `Sdf.Path.HasPrefix`, which compares path elements. Depth is counted in path elements below the root (`pathElementCount`) instead of splitting the string on `/`. A missing root raises the same error as before.

### Verified on 22.0.429

Three `primitive` LOPs (`/materials/BLD`, `/materials/BLD_probes`, `/materials/BLD/wood`): `root_path=/materials/BLD` → `count: 2` (`BLD`, `BLD/wood`), `BLD_probes` absent; `root_path=/materials, depth=1` → 3; `root_path=/` → 5 as before. `get_usd_prim_stats(prim_path=/materials/BLD)` → `total_prims: 2` (was 3).

### Tests

`tests/test_list_usd_prims_root.py`: a sibling sharing the prefix is not listed, `/` still lists everything, depth counts elements below the root, prim stats use the same test. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: 0ab9384. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #75, opened 2026-09-15
list_hda_versions lists the whole version family of a type

d-potrivaev:
`list_hda_versions` iterates `node.type().allInstalledDefinitions()`. Houdini keeps an asset's version in the type name (`brick::1.1` is a different `NodeType` from `brick`), so that lists the definitions of one type across files and never the version family the tool's name promises: after copying a definition to `brick::1.1` and installing it, the reply still showed one row, `1.0`, and "is `::1.1` installed?" had to be answered with `hou.nodeType()` by hand.

### Change

The type name is split with `hou.hda.componentsFromFullNodeTypeName`; the category's `nodeTypes()` are walked for every type with the same scope, namespace and name, and each installed definition becomes a row with `type_name`, `version` (the definition's, else the one in the name), `library_file`, `is_current`, `is_preferred` and `is_instance_type` (whether it is the type of the node asked about). Rows sort numerically (`1.2` before `1.10`, unversioned first). The reply carries `family: {scope, namespace, name}`.

### Verified on 22.0.429

`brick` (1.0) plus a copy installed as `brick::1.1` → two rows: `(brick, 1.0, current, instance type)` and `(brick::1.1, 1.1, current, not the instance type)`. A stale definition of the same name left in another file showed up too, which is the point.

### Tests

`tests/test_hda_versions.py`: every version of the name is listed with the right flags and a look-alike name is not, versions sort numerically. Unit suite passes, ruff clean.


healkeiser:

Your work is on `main`, cherry-picked with your authorship intact: a1e924a. The only edit is a conventional prefix on the subject line; the diff and your commit body are untouched. Lint, the 567 unit tests and the generated-content check all pass on the result.

PR #76, opened 2026-09-16
feat(graph): get_node_card lists connectors by index and name, and build_network wires inputs by input_name

d-potrivaev:
## What this changes

The card reported min/max input counts and nothing about the connectors
themselves, so the index of `texcoord` on mtlximage (3) could not be read
from any tool; and a build_network spec named an input by index only, while
connect_nodes had taken an input_name for months. A session wiring a
MaterialX texture had to guess the index and read the error.

get_node_card now carries `inputs` and `outputs`, each entry with index,
name and label (and the VOP data type where Houdini exposes one). Connector
names live only on instances — hou.NodeType has no inputNames — so the card
makes a probe inside a throwaway container under /obj (geo, matnet,
lopnet, ... per category) and destroys it in the same call; measured at
1-6 ms per card on 22.0.429, karma inside a ropnet the slow one at 57 ms.
build_network's probe pass fills the same table, so a spec entry may say
{"source": "img", "input_name": "base_color"}: resolved at dry-run time,
and a wrong name fails validation with a did-you-mean over names and labels
("base_colour" -> Did you mean: ['base_color', 'sheen_color',
'coat_color']?) instead of the build. The client tool gains
include_help=False for callers who only want names or connectors.

Live on 22.0.429: mtlximage lists texcoord as input 3 labelled "Texture
Coordinates"; polyextrude lists "Geometry for operation" / "Spine curve";
a network built with input_name="base_color" lands the image on input 1 of
mtlxstandard_surface; the probe container is gone after the call.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 589 passed (`tests/test_node_card_connectors.py` and the updated neighbours)
- [x] Generated files regenerated with their `tools/` script, not by hand (`gen_required_commands.py` run, no change (no new commands))

Updated after review in dc8f4fa.


houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+    """
+    container = _SCRATCH_CONTAINERS.get(category_name)
+    if container is None:
+        return None
+    root = hou.node("/obj")
+    if root is None:
+        return None
+    scratch = None
+    try:
+        scratch = root.createNode(container, "fxhoudinimcp_card_probe")
```
healkeiser: **get_node_card, a read-only documentation command, now creates and destroys a node under /obj on every call, so it dirties the hip file and burns one undo-stack slot per card.**

dispatcher.py:211 wraps every command in `_undo_group(command)`, and the docstring at dispatcher.py:104 relies on 'Houdini discards a block that recorded nothing' to keep read-only commands off the undo stack. The probe now records createNode + destroy, so the block is kept. A session calls get_node_card 8 times (the shipped prompt at prompts/markdown/instructions/discipline.md tells the model to look up every type before use), the user presses Ctrl+Z to undo their own last edit and gets 8 no-op 'MCP graph.get_node_card' steps first; hou.hipFile.hasUnsavedChanges() is now True on a scene nobody edited, and real undo history is pushed out of the buffer. Wrap the probe in `hou.undos.disabler()`.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -683,13 +815,19 @@ def _collect_multiparms(entries) -> None:
     if help_text and len(help_text) > 5000:
         help_text = help_text[:5000] + "\n[... help truncated]"
 
+    # Connectors, in order, by index AND name: the index of `texcoord` on
+    # mtlximage is 3, and until now there was nowhere to read that.
+    connectors = _connectors_for_type(context, resolved) or {"inputs": [], "outputs": []}
```
healkeiser: **`_connectors_for_type` returns None to mean 'could not probe this category', and the call site immediately collapses that into empty lists, so the card reports 'this type has no connectors' when it actually means 'I did not look'.**

`_connectors_for_type` docstring says None means 'the category has no known container (nothing to probe in)'. `connectors = _connectors_for_type(context, resolved) or {"inputs": [], "outputs": []}` throws that signal away. get_node_card("vopcop2gen", context="Cop2") - a category that exists in hou.nodeTypeCategories() on every supported 20.5-22.0 build and that cop_handlers.py:393 explicitly still falls back to - is not in _SCRATCH_CONTAINERS, so the card returns `"inputs": []` alongside `"max_inputs": 2`. Same for "Shop", "VopNet", "ChopNet", and for any type whose probe raised (license, container mismatch). A caller reads `inputs: []` as authoritative and wires nothing. Omit the keys, or add `"connectors_probed": false`.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -411,12 +540,15 @@ def build_network(
...
             for input_index, entry in enumerate(spec.get("inputs") or []):
+                input_name = None
                 if isinstance(entry, dict):
                     source_name = entry.get("source")
+                    input_name = entry.get("input_name")
                     input_index = int(entry.get("index", input_index))
                     source_output = int(entry.get("source_output", 0))
                 else:
                     source_name, source_output = entry, 0
+                input_index = _resolve_input_index(node, input_index, input_name)
```
healkeiser, line 592: **dry_run validates `input_name` against a fresh default-parm probe, but the build resolves the same name on the node after its parms are set, so a dry run can pass and the real build still raise and roll the whole network back.**

Phase 2 applies `parms` (line ~528) and only then calls `_resolve_input_index(node, input_index, input_name)` on the configured node. For any type whose connector set depends on a parameter - the VOP `signature` parm that `_is_dynamic_menu` at line 94 documents as changing what the node exposes, a multiparm-driven HDA, a variadic SOP - the probe's names and the live node's names differ. Spec `{"type": "mtlxswitch", "parms": {"signature": "color3"}, "inputs": [{"source": "a", "input_name": "in3"}]}`: dry_run returns valid:true, the build raises ValueError from _resolve_input_index, the `except Exception` at line 565 destroys every node created so far and returns 'build failed and was rolled back'. The whole point of the two-phase design is that a passing dry run means the build will work.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+    """Input and output connectors of a live node, in order, with their names.
+
+    A VOP's inputs are addressed by name (`texcoord`), and their order is not
+    stable across versions; a SOP's are `input1`, `input2` behind labels such
+    as "Spine curve". Both the index and the name are reported so a caller can
+    wire by whichever the verb at hand accepts.
+    """
+    inputs: list[dict[str, Any]] = []
+    with contextlib.suppress(Exception):
+        names = list(node.inputNames())
```
healkeiser: **`_connectors_of` reads the connector list off an unconnected probe, so types with variadic inputs report only the connectors that exist at zero connections - the card understates them and dry-run rejects legitimate names.**

hou.Node.inputNames() on a live node returns the connectors the node currently shows, and a merge/switch/object_merge SOP grows its input list as inputs are attached. A fresh `merge` probe reports one connector, so get_node_card('merge') returns `max_inputs: 9999` next to `inputs: [{index: 0, name: 'input1'}]`. Worse, build_network with `{"type": "merge", "inputs": [{"source": "a", "input_name": "input1"}, {"source": "b", "input_name": "input2"}]}` fails validation at line 481 with "no input named 'input2' on merge" for a wiring that is perfectly legal - the index form of the same spec builds fine.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -353,10 +467,25 @@ def build_network(
                     )
         node_type = resolved_types.get(spec.get("type"))
         max_inputs = node_type.maxNumInputs() if node_type else 0
+        connectors = connector_knowledge.get(spec.get("type"))
```
healkeiser: **When the node type failed to resolve there is no connector entry, and the `input_name` branch emits a bogus 'no input named X' error on top of the real 'type does not exist' error - the parm loop guards this case, the new one does not.**

The parm validation at line 447 is guarded by `if knowledge:` precisely so an unresolved type produces one error, not a cascade. `connectors = connector_knowledge.get(spec.get('type'))` returns None for the same unresolved type, `_connector_index(None, ...)` returns None, and `_connector_hint(None, ...)` returns "". Spec `[{"type": "mtlxstandard_surfce", "inputs": [{"source": "img", "input_name": "base_color"}]}]` returns both "type 'mtlxstandard_surfce' does not exist" and "no input named 'base_color' on mtlxstandard_surfce." - the second is noise that sends the caller hunting a connector problem that does not exist. Skip the input_name check when `connectors is None`, mirroring `if knowledge:`.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -353,10 +467,25 @@ def build_network(
...
+                    # A connector named by the caller, as the node card lists
+                    # it. Resolved here so a wrong name fails the dry run,
+                    # not the build.
+                    found = _connector_index(connectors, str(input_name))
+                    if found is None:
+                        errors.append(
+                            f"node {label}: no input named '{input_name}' on "
+                            f"{spec.get('type')}.{_connector_hint(connectors, str(input_name))}"
+                        )
+                        continue
```
healkeiser: **The `continue` after an unresolved input_name skips the source-existence check for that entry, so build_network stops reporting every problem in one pass.**

The comment at line 419 states the design: 'Runs even when other errors exist: report everything in one pass.' Spec `{"source": "imge", "input_name": "base_colour"}` has two typos. The first dry run reports only the connector name; the caller fixes it and the second dry run reports the bad source. Two round trips where the file's own contract promises one. Append the error and fall through to the source check instead of `continue`.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+    container = _SCRATCH_CONTAINERS.get(category_name)
+    if container is None:
+        return None
+    root = hou.node("/obj")
+    if root is None:
+        return None
+    scratch = None
+    try:
+        scratch = root.createNode(container, "fxhoudinimcp_card_probe")
+        probe = scratch.createNode(node_type.name())
```
healkeiser: **The probe is created with createNode's default run_init_scripts=True, so a pure documentation read executes the node type's creation scripts.**

get_node_card('myStudio::rigLoader::2.0', context='Object') on a studio HDA runs the asset's OnCreated / creation script as a side effect of asking for its parameter list - that script can write files, hit a license server, pop a dialog that blocks the main thread past the command timeout, or mutate the scene in ways the destroy() does not undo. Pass `run_init_scripts=False` to both probe creations (line 197 and _parm_names_for_type line 221); parm names, connector names and labels all come from the type, not from the init script.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+        labels = list(node.outputLabels())
+        for index, name in enumerate(names):
+            entry = {"index": index, "name": name}
+            if index < len(labels):
+                entry["label"] = labels[index]
+            outputs.append(entry)
+    return {"inputs": inputs, "outputs": outputs}
+
+
+def _connector_index(connectors: dict | None, input_name: str) -> int | None:
```
healkeiser: **`_connector_index` / `_connector_hint` re-implement the name-or-label lookup and did-you-mean that `node_handlers._resolve_input_index` already does - and this PR imports that very function - and the two resolution rules already disagree.**

node_handlers.py:551-557 checks every *name* first, then every *label*, then builds the hint from `names + labels`. `_connector_index` scans entries in order accepting name-or-label per entry. On a type where input 0's label equals input 2's name, dry-run resolves the string to 0 and the build (via _resolve_input_index at line 551) resolves it to 2, so the network wires to a different connector than the one validated. Two copies of one rule that must agree, guaranteed to drift. Give `_resolve_input_index` an optional pre-built connector table and call the one function from both phases.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+    close = get_close_matches(input_name, candidates, n=3, cutoff=0.4)
+    return f" Did you mean: {close}?" if close else f" Inputs: {candidates}"
+
+
+# OBJ-level container whose children belong to each node category. A card is
+# asked for a TYPE, and connector names live only on instances (hou.NodeType
+# has no inputNames), so a probe is made inside a throwaway container and
+# destroyed again. Measured at 1-6 ms per card; karma inside a ropnet is the
+# slow one at 57 ms.
+_SCRATCH_CONTAINERS = {
```
healkeiser: **`_SCRATCH_CONTAINERS` hardcodes nine category-to-container pairs instead of deriving the container from `hou.NodeType.childTypeCategory()`, so every category SideFX has or adds that is not in the list silently loses its connectors.**

The map is a special case layered on a mechanism Houdini already exposes: scanning `hou.objNodeTypeCategory().nodeTypes()` once for a type whose `childTypeCategory().name()` matches gives geo/subnet/matnet/lopnet/dopnet/copnet/chopnet/topnet/ropnet *and* cop2net, shopnet and anything future, with no table to maintain. As written, 'Cop2' and 'Shop' are already missing on the 20.5-22.0 range the README supports, and 'Vop' is pinned to `matnet`, a type SideFX has been deprecating - the day it goes, every VOP card silently returns empty connectors through the `except Exception: return None` at line 199 with no diagnostic.

python/fxhoudinimcp/tools/graph.py
```diff
@@ -100,11 +104,15 @@ async def get_node_card(
         context: Category — "Sop", "Lop", "Dop", "Cop", "Chop", "Top",
```
healkeiser: **The client tool's `context` docstring lists eight categories and omits "Vop" - the one category the whole PR exists to serve.**

The PR's motivating example is reading the index of `texcoord` on mtlximage, a Vop type, and `_SCRATCH_CONTAINERS` maps "Vop" -> "matnet" for exactly that. But the only documentation the model sees says: context: Category - "Sop", "Lop", "Dop", "Cop", "Chop", "Top", "Object", "Driver". A caller wanting the mtlximage card has no listed context that resolves it; get_node_card('mtlximage', 'Sop') raises "Node type 'mtlximage' not found in Sop". Add "Vop" to the list.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+def _connectors_of(node: hou.Node) -> dict[str, list[dict[str, Any]]]:
+    """Input and output connectors of a live node, in order, with their names.
+
+    A VOP's inputs are addressed by name (`texcoord`), and their order is not
+    stable across versions; a SOP's are `input1`, `input2` behind labels such
+    as "Spine curve". Both the index and the name are reported so a caller can
+    wire by whichever the verb at hand accepts.
+    """
+    inputs: list[dict[str, Any]] = []
+    with contextlib.suppress(Exception):
```
healkeiser: **`inputNames()` and `inputLabels()` share one `contextlib.suppress` block, so a failure in the labels call throws away the names too - unlike `inputDataTypes()`, which is correctly nested in its own suppress.**

The author already recognised the pattern one line down: `data_types` gets its own nested suppress and the test `test_a_node_without_data_types_still_lists_its_inputs` proves inputs survive that failure. Give `inputLabels()` the same treatment. If `node.inputLabels()` raises on any type (or returns a non-iterable), the whole `for index, name in enumerate(names)` loop is skipped and `inputs` stays `[]`, which the card then reports as 'this type has no inputs' - the same silent lie as the `_connectors_for_type` finding, from a different cause.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+        return _connectors_of(probe)
+    except Exception:
+        return None
+    finally:
+        with contextlib.suppress(Exception):
+            if scratch is not None:
+                scratch.destroy()
+
+
+def _parm_names_for_type(
```
healkeiser: **`_parm_names_for_type` takes a caller-allocated dict as an out-parameter and mutates it, when it could simply return the connectors as a fifth element or the caller could call `_connectors_of` itself.**

The call site needs three lines (`connector_knowledge[type_name] = {}` then pass it in then read it back) where one would do, and the function now has two output channels: a return tuple and a side effect on an argument. The tuple return type annotation `tuple[set, set, dict, list]` no longer describes everything the function produces, and every reader has to check whether a given call site passed the dict. Return `tuple[set, set, dict, list, dict]`, or drop the parameter and have build_network call `_connectors_of` on its own probe.

python/fxhoudinimcp/tools/graph.py
```diff
@@ -86,10 +87,13 @@ async def get_node_card(
     node_type: str,
     context: str = "Sop",
     parm_filter: str | None = None,
+    include_help: bool = True,
```
healkeiser, line 90: **The PR adds `include_help=False` to make a card cheaper while making every card pay an unconditional scene mutation with no opt-out and no memoisation.**

A caller who passes include_help=False to save ~4 KB still pays the probe: container createNode + type createNode + destroy, which the PR itself measures at 1-6 ms and 57 ms for karma in a ropnet, plus the undo entry from the first finding - and pays it again on every repeat lookup of the same type in the same session, because nothing caches the result. Either memoise by (context, resolved.name()) for the process, or add the symmetric `include_connectors` flag so a caller who only wants parm names can skip it.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -102,16 +103,123 @@ def _is_dynamic_menu(probe: hou.Node, parm: hou.Parm) -> bool:
...
+    if not connectors:
+        return ""
+    candidates = [
+        value
+        for entry in connectors.get("inputs", [])
+        for value in (entry.get("name"), entry.get("label"))
+        if value
+    ]
+    close = get_close_matches(input_name, candidates, n=3, cutoff=0.4)
+    return f" Did you mean: {close}?" if close else f" Inputs: {candidates}"
```
healkeiser: **`_connector_hint`'s no-close-match fallback dumps every connector name AND every label into the error string.**

`f" Inputs: {candidates}"` where candidates is names + labels interleaved, one pair per input. On mtlxstandard_surface (roughly 30 inputs) a single typo produces an error line of ~60 quoted strings, duplicated per offending entry, in a response the model has to read. Cap the list the way `_menu_error` already does at line 274 (`tokens[:15] + ["..."]`), and dedupe names against labels.

healkeiser (commented):
### Code review

14 findings. 14 are inline below.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed what can be in dc8f4fa: the probe runs under `hou.undos.disabler()` (measured: no undo entry), and its result is cached per type for the session. The modified flag cannot be avoided — Houdini sets it on createNode even under the disabler, HOM has no call to clear it, and `hou.NodeType` has no `inputNames` to read instead — so the first card for a type marks the scene modified once, as the build_network dry run's probe already did. The docstring says so.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: when connectors could not be read the card has `connectors_probed: false` and `connectors_note` with the reason, and no `inputs`/`outputs` keys.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: a spec that names an input and sets parms is probed with those parms applied, so validation resolves against the connectors the build will see. (On 22.0.429 `mtlxswitch`'s `signature` does not rename its inputs, but the rule covers any type whose connectors depend on a parameter; unit test with a probe that grows an input.)

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Confirmed live: a fresh merge lists `input1` only and grows as inputs are wired (max 9999). Fixed in dc8f4fa: a name further along the same numbered series resolves (`input2` -> 1) within `maxNumInputs`; live test wires a merge by `input1`/`input2` through build_network.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: no connector check when the type did not resolve.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: the connector error is appended and the source is still checked; a unit test gets both in one dry run.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed for get_node_card in dc8f4fa (`run_init_scripts=False`). build_network's probe keeps running them on purpose: the build that follows runs them, so the parm names it validates are the ones the built node will have.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: `node_handlers._find_input` (names, then labels, then a variadic series) is the one rule; the dry run and `_resolve_input_index` both call it.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: any category without a preferred container is found through `childTypeCategory()` (non-hidden first). Live: a `blend` card in Cop2 (cop2net) lists its inputs.

python/fxhoudinimcp/tools/graph.py
d-potrivaev: Fixed in dc8f4fa: Vop, plus Cop2, Shop, VopNet.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: names, labels and data types are each read under their own guard.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: `_parm_names_for_type` returns the connectors as a fifth element.

python/fxhoudinimcp/tools/graph.py
d-potrivaev: Fixed in dc8f4fa: cached per (category, type, HDA definition modification time) for the session.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in dc8f4fa: capped at 15, names and labels deduplicated.

d-potrivaev:
Thanks for the review. All 14 findings are answered in dc8f4fa, one commit on top of the branch: 14 addressed as suggested. Each inline thread has its answer.

Verified: ruff check + format, `pytest -m "not integration"` (589 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_graph_live.py::TestNodeCardConnectors) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: findings are fixed or correctly rebutted, unit tests pass (589).

PR #77, opened 2026-09-16
feat(nodes): a subnet's indirect inputs are a wiring source for connect_nodes, connect_nodes_batch and build_network

d-potrivaev:
## What this changes

A subnet's input connectors are SubnetIndirectInput items, not nodes: they
have no path, and every wiring verb names its source by path. So the first
node of a chain built inside a subnet could not be fed from the subnet's own
input by any tool, and a session ended up in execute_python with
node.setInput(0, subnet.indirectInputs()[0]).

connect_nodes and connect_nodes_batch take indirect_input=n with the subnet
as source_path; build_network takes {"indirect_input": n} in place of a
source. An index past the subnet's connectors, or a plain node given as the
subnet, is refused by name ("/obj/geo1/subnet1 has 4 input connector(s);
asked for #9"), at dry-run time for build_network. The reply keeps
source_path as the subnet's path and names the connector in indirect_input.

Live on 22.0.429: a matchsize built inside an empty subnet with
{"indirect_input": 0} is wired to an OpSubnetIndirectInput. Note that
hou.Node.inputs() lists nodes only and shows nothing for such a wire; it is
visible through inputConnections()[i].inputItem().

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 590 passed (`tests/test_subnet_indirect_inputs.py` and the updated neighbours)
- [x] Generated files regenerated with their `tools/` script, not by hand (`gen_required_commands.py` run, no change (no new commands))

Updated after review in 61c9cf3.


houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -559,37 +559,68 @@ def _resolve_input_index(dest: hou.Node, input_index: int, input_name: str | Non
...
+        items = list(node.indirectInputs())
+    except Exception as exc:
+        raise ValueError(
+            f"{source_path} has no indirect inputs (not a subnet): {readable_message(exc)}"
+        ) from exc
+    if not 0 <= int(indirect_input) < len(items):
+        raise ValueError(
+            f"{source_path} has {len(items)} indirect input(s); asked for #{indirect_input}."
+        )
+    return items[int(indirect_input)], f"{node.path()} (indirect input {int(indirect_input)})"
```
healkeiser: **_resolve_source never checks that dest lives inside the subnet, so the documented precondition surfaces as a raw hou.OperationFailed instead of the named refusal the PR promises.**

connect_nodes('/obj/geo1/subnet1', '/obj/geo1/box1', indirect_input=0) - dest outside the subnet. The docstring says "the node at dest_path must live inside that subnet" but nothing enforces it; dest.setInput() then raises hou.OperationFailed ("item_to_become_input is not in the same network as this node", per the HOM setInput docs), which is exactly the opaque failure the PR's other two refusals were written to replace. One check in _resolve_source (dest.parent() is the subnet) would name it; the helper does not currently receive dest at all.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -411,17 +437,22 @@ def build_network(
...
                     source_output = int(entry.get("source_output", 0))
                 else:
                     source_name, source_output = entry, 0
-                source = (
-                    created.get(source_name)
-                    or parent.node(str(source_name))
-                    or hou.node(str(source_name))
-                )
+                if indirect is not None:
+                    source = list(parent.indirectInputs())[int(indirect)]
```
healkeiser: **build_network's actual indirect wiring line is untested - both new build_network tests pass dry_run=True, so Phase 2 never runs.**

test_a_connector_index_validates_without_a_source and test_an_index_past_the_connectors_fails_the_dry_run both return before Phase 2. `source = list(parent.indirectInputs())[int(indirect)]` and the following `node.setInput(input_index, source, source_output)` are never executed by any test; a wrong index expression, a missed source_output, or an exception type change here ships green. The connect_nodes tests cover a different code path (node_handlers._resolve_source), so build_network's copy has zero coverage on the only line that mutates the scene.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -559,37 +559,68 @@ def _resolve_input_index(dest: hou.Node, input_index: int, input_name: str | Non
...
+    node = _get_node(source_path)
+    if indirect_input is None:
+        return node, node.path()
+    try:
+        items = list(node.indirectInputs())
+    except Exception as exc:
+        raise ValueError(
+            f"{source_path} has no indirect inputs (not a subnet): {readable_message(exc)}"
+        ) from exc
+    if not 0 <= int(indirect_input) < len(items):
```
healkeiser: **_resolve_source lets a non-integer indirect_input escape as a bare int() ValueError, while the graph-side twin returns a named message - and silently truncates floats.**

connect_nodes_batch([{"source_path": subnet, "dest_path": dest, "indirect_input": "first"}]) - `connections` is typed list[dict[str, Any]] so the MCP schema does not guard the value. int("first") raises "invalid literal for int() with base 10: 'first'", which is Python talking about itself, the exact failure mode dispatcher._argument_error was written to avoid. graph_handlers._indirect_input_error (line 189-191) handles the same input with "indirect_input must be an integer, got 'first'". Conversely indirect_input=1.9 passes and silently wires connector 1.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -364,6 +383,13 @@ def build_network(
                     f"node {label}: input {input_index} exceeds max inputs "
                     f"({max_inputs}) of {spec.get('type')}"
                 )
+            if indirect is not None:
+                # The parent subnet's own connector: not a node, so it has no
+                # path a source string could name.
+                problem = _indirect_input_error(parent, indirect)
+                if problem:
+                    errors.append(f"node {label}: {problem}")
+                continue
```
healkeiser, line 412: **build_network silently ignores `source` when `indirect_input` is also present, and the `continue` makes a bogus source pass the dry run.**

{"type": "matchsize", "inputs": [{"source": "doesnotexist", "indirect_input": 0}]} validates clean (valid: True) and builds wired to the connector. The caller asked for two mutually exclusive sources and gets no word about it; a typo'd or stale `source` left in a spec alongside a newly added indirect_input is swallowed rather than reported, which is the opposite of the upfront-validation contract this function advertises. Two lines: error out when both keys are set.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -559,37 +559,68 @@ def _resolve_input_index(dest: hou.Node, input_index: int, input_name: str | Non
...
     input_index = _resolve_input_index(dest, input_index, input_name)
     dest.setInput(input_index, source, output_index)
 
     _focus_network_editor(dest, place_unpositioned=False)
 
-    return {
+    result = {
         "success": True,
-        "source_path": source.path(),
+        "source_path": source_label,
```
healkeiser: **connect_nodes now returns a non-path string in the field named source_path, breaking the field's contract for the agent that reads the reply.**

connect_nodes(subnet, dest, indirect_input=0) returns {"source_path": "/obj/geo1/subnet1 (indirect input 0)"}. An agent chaining calls - feeding the reported source_path into the next connect_nodes or get_node_info - gets "Node not found: /obj/geo1/subnet1 (indirect input 0)". The reply already carries a separate `indirect_input` key that disambiguates, so source_path could stay a real path and the prose go in its own key (e.g. `source`).

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -411,17 +437,22 @@ def build_network(
...
-                )
+                if indirect is not None:
+                    source = list(parent.indirectInputs())[int(indirect)]
+                else:
+                    source = (
+                        created.get(source_name)
+                        or parent.node(str(source_name))
+                        or hou.node(str(source_name))
+                    )
                 node.setInput(input_index, source, source_output)
```
healkeiser, line 471: **build_network does not validate source_output for an indirect input, so a bad value passes the dry run and then rolls the whole build back at Phase 2.**

{"indirect_input": 0, "source_output": 2}: a hou.SubnetIndirectInput has exactly one output, so setInput raises hou.InvalidInput ("Raises hou.InvalidInput if output_index is invalid", HOM setInput). Phase 1 continues past every source check for indirect entries, so dry_run reports valid: True; the real run destroys every node it created and returns "build failed and was rolled back". The function's stated promise is that everything in the spec is validated before anything is created.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -179,6 +179,21 @@ def _apply_parm(node: hou.Node, name: str, value: Any) -> None:
         raise ValueError(f"parameter '{name}' not found")
 
 
+def _indirect_input_error(parent: hou.Node, indirect: Any) -> str | None:
```
healkeiser: **The same indirect-input rule is implemented twice, in two modules, with divergent messages and divergent behaviour on a non-integer.**

_indirect_input_error (graph_handlers:182) and _resolve_source (node_handlers:562) both answer "is connector n of this subnet wirable?" - but one returns a string and one raises, one names the non-integer case and one throws int()'s message, one appends readable_message(exc) to the 'not a subnet' text and one does not. A user hitting the same mistake through build_network and through connect_nodes gets two different answers, and any future fix has to be made twice. One shared helper (config.py already hosts cross-handler helpers) returning the item or raising would cover both.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -354,8 +371,10 @@ def build_network(
         node_type = resolved_types.get(spec.get("type"))
         max_inputs = node_type.maxNumInputs() if node_type else 0
         for input_index, entry in enumerate(spec.get("inputs") or []):
+            indirect = None
```
healkeiser: **The input-entry parsing is now duplicated between Phase 1 and Phase 2 with a deliberate asymmetry, so the two phases can disagree about what a spec means.**

Lines 373-380 and 440-447 both decode {index, source, indirect_input} from the entry, but only Phase 2 reads source_output - which is precisely why the unvalidated source_output finding exists. Every future input key has to be added in both places or validation and execution drift apart. One `_parse_input_entry(entry, position)` returning (index, source, indirect, source_output) removes the class of bug rather than this instance of it.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -179,6 +179,21 @@ def _apply_parm(node: hou.Node, name: str, value: Any) -> None:
         raise ValueError(f"parameter '{name}' not found")
 
 
+def _indirect_input_error(parent: hou.Node, indirect: Any) -> str | None:
+    """Why connector *indirect* of *parent* cannot be wired from, or None."""
+    try:
+        count = len(list(parent.indirectInputs()))
+    except Exception:
```
healkeiser: **The bare `except Exception` reports "not a subnet" for any failure of indirectInputs(), including causes that have nothing to do with the node type.**

A subnet inside a locked HDA, or any other HOM error from indirectInputs(), is reported to the user as "/obj/geo1/subnet1 has no indirect inputs (not a subnet)" - a confidently wrong diagnosis that sends them looking at the wrong thing. The node-side twin at least appends readable_message(exc); this one discards the exception entirely. Either narrow to hou.InvalidNodeType (what the HOM docs say is raised) or carry the message.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
```diff
@@ -179,6 +179,21 @@ def _apply_parm(node: hou.Node, name: str, value: Any) -> None:
         raise ValueError(f"parameter '{name}' not found")
 
 
+def _indirect_input_error(parent: hou.Node, indirect: Any) -> str | None:
+    """Why connector *indirect* of *parent* cannot be wired from, or None."""
+    try:
+        count = len(list(parent.indirectInputs()))
```
healkeiser: **parent.indirectInputs() is re-listed once per input entry in Phase 1 and again once per entry in Phase 2, instead of once per build.**

A subnet spec with N indirect-fed inputs costs 2N main-thread HOM round trips to fetch a tuple that cannot change during the call - and the module's own dispatcher docstring prices a main-thread marshal at ~50 ms. Hoist `connectors = list(parent.indirectInputs())` once before the spec loop and index it in both phases.

python/fxhoudinimcp/tools/nodes.py
```diff
@@ -237,29 +237,37 @@ async def connect_nodes(
...
-            "input_name": input_name,
-        },
-    )
+    params: dict[str, Any] = {
+        "source_path": source_path,
+        "dest_path": dest_path,
+        "output_index": output_index,
+        "input_index": input_index,
+        "input_name": input_name,
+    }
```
healkeiser, line 268: **The compat manifest only tracks command names, so a plugin older than the server reports "compatible" and then fails confusingly on the new indirect_input parameter.**

The two halves ship separately (compat.py: "Upgrading one and not the other is easy to do and, until now, invisible"), and required_commands() compares names only - nodes.connect_nodes exists on an old plugin, so missing_commands() stays empty. connect_nodes(..., indirect_input=0) then reaches dispatch's handler(**params) and returns BAD_ARGUMENTS about an unexpected keyword. Worse in the batch path: connect_nodes_batch passes the dict straight through, an old handler never reads the key, and the call silently tries to wire the subnet node itself as the source. The client already omits the key when None, which limits this to users of the new feature, but nothing tells them the plugin is the reason.

healkeiser (commented):
### Code review

13 findings. 11 are inline below. 2 land outside this diff and are listed here:

- `houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py`:754 - reorder_inputs destroys an indirect-input wire and then either fails or silently drops it, because node.inputs() hides the SubnetIndirectInput and resolves it to a node outside the network.

  Wire /obj/geo1/subnet1/matchsize1 from connector 0 with the new connect_nodes(indirect_input=0), then call reorder_inputs('/obj/geo1/subnet1/matchsize1', [0]). HOM (hou.Node.inputs, Houdini 22 help/hom.zip): "If an input is connected to a hou.SubnetIndirectInput, the node connected to the corresponding input on the parent subnet is returned... the presence of the indirect input is hidden." So current_inputs[0] is either None (subnet input unconnected -> the disconnect loop at line 749 calls setInput(0, None), killing the wire, and the reconnect loop skips it: wire gone, success:True returned) or the external upstream node (subnet input connected -> setInput(new_idx, external_node) raises hou.OperationFailed "item_to_become_input is not in the same network" AFTER everything was already disconnected). Either way the node is left unwired. The PR makes this wire creatable through a supported verb without teaching reorder_inputs about inputConnections()/inputItem().

- `houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py`:703 - disconnect_node cannot remove an indirect-input wire and reports success (or a wrong error) instead, so the new capability creates wires no verb can undo.

  Exactly the PR's headline case: a matchsize built inside an empty subnet, wired with indirect_input=0. The subnet's own input 0 is unconnected, so node.inputs() yields () or (None,). disconnect_node(path, disconnect_all=True) then iterates nothing (or skips the None) and returns {"success": True, "disconnected_inputs": []} while the wire is still there; disconnect_node(path, input_index=0) raises "Input index 0 is out of range or already disconnected" for an input that is plainly connected in the network editor. The only way back out is execute_python - the very thing this PR exists to avoid.


houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in 61c9cf3: refused by name ("... is not inside <subnet>: a subnet's input connector feeds only nodes inside that subnet") before setInput, which raised a bare OperationFailed.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in 61c9cf3: a unit test runs the build and asserts `setInput(1, connector, 0)`; a live test builds from the connector.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in 61c9cf3: one helper refuses a non-integer by name; an integral float is accepted, 1.9 is refused.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in 61c9cf3: both together fail the dry run.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in 61c9cf3: `source_path` is the subnet's path; the connector is in `indirect_input`.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Confirmed live (`InvalidInput`). Fixed in 61c9cf3: a non-zero source_output on a connector fails the dry run; connect_nodes refuses output_index != 0 by name.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in 61c9cf3: `node_handlers._indirect_input_item` is the one rule, used by all three verbs.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in 61c9cf3: `_parse_input_entry` feeds both phases.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Narrowing to `InvalidNodeType` would not help: on 22.0.429 a Box answers `indirectInputs()` with an empty tuple rather than raising. 61c9cf3 refuses an empty list as "not a subnet, or a subnet without inputs" and carries the exception's own message when one is raised.

houdini/scripts/python/fxhoudinimcp_server/handlers/graph_handlers.py
d-potrivaev: Fixed in 61c9cf3: listed once per build.

python/fxhoudinimcp/tools/nodes.py
d-potrivaev: Measured on 22.0.429: on a plugin without this change a batch entry does not wire the subnet — `setInput` refuses the subnet itself (`OperationFailed`), so it lands in `errors`. connect_nodes gets BAD_ARGUMENTS naming `indirect_input`, and in 61c9cf3 the client adds that the plugin predates the parameter. Checking parameters in the manifest would be a change to compat for every command; happy to do that separately if you want it.

d-potrivaev:
Thanks for the review. All 13 findings are answered in 61c9cf3, one commit on top of the branch: 11 addressed as suggested, 2 addressed differently or in part. Each inline thread has its answer.

Findings outside the diff:
- **reorder_inputs (node_handlers.py:754).** Confirmed live: `inputs()` hides the wire. Fixed in 61c9cf3: reorder_inputs reads `inputConnections()` (item and its output index per wire), checks the whole order before disconnecting anything, and puts the connector back; live test `test_reorder_keeps_the_connector`.
- **disconnect_node (node_handlers.py:703).** Fixed in 61c9cf3: disconnect_node works from `inputConnections()`, so the wire this PR makes can be removed (live test `test_the_wire_is_made_and_can_be_removed`).

Verified: ruff check + format, `pytest -m "not integration"` (590 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_nodes_live.py::TestSubnetInputConnectors) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: findings are fixed or correctly rebutted, unit tests pass (590). The one leftover (build_network swallowing the indirectInputs() error) was fixed on main in 41293d6.

PR #78, opened 2026-09-16
feat(parameters): get_parm_references and get_parm_template_tree read who references a parameter and the whole interface as a tree

d-potrivaev:
## What this changes

Two questions a live session kept answering with execute_python.

"Who reads this parameter, and what does it read" had no tool: renaming a
control on an asset meant a HOM script over parmsReferencingThis before
anything could be trusted. get_parm_references answers both directions in
one call -- `incoming` per parameter (the parameters elsewhere whose
expressions read it), `outgoing` (what this node's expressions read,
resolved to paths), plus the node-level dependents and references.
getReferencedParm() resolves only a pure ch("../src/tx"); for
ch("../src/scale") * 2 it answers with the parm itself (measured on
22.0.429), so richer expressions are parsed for ch/chs/chf references and
resolved relative to the node, and the reply marks `pure_reference`.

"What is in the Controls tab, in order, with its Hide When rules" had no
answer either: get_parameter_schema flattens folders, conditionals, menu
items and multiparm blocks away, and get_hda_info shows only the top
folders. get_parm_template_tree returns the interface as Type Properties
shows it -- folders with their folder_type, every parameter in order with
defaults, default expressions, ranges, menu items with their labels,
conditionals, callbacks and naming scheme -- for a node or for a node type
(resolved through the preferred version, as createNode does). `folder`
narrows to one folder by label; `max_entries` cuts depth-first and the
reply says so.

Live on 22.0.429: incoming on a null names both readers of tx and scale;
outgoing on the reading node gives the pure reference and the parsed one;
the polyextrude::2.0 tree is 102 entries under Extrusion, Spine Control
and Local Control, with splittype's menu carrying value and label.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 591 passed (`tests/test_parm_references.py` and the updated neighbours)

Updated after review in 50b0725.


houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+    with contextlib.suppress(Exception):
+        components = pt.numComponents()
+        if components != 1:
+            entry["components"] = components
+    with contextlib.suppress(Exception):
+        scheme = pt.namingScheme().name()
+        if scheme != "XYZW" or entry.get("components", 1) > 1:
+            entry["naming_scheme"] = scheme
+    with contextlib.suppress(Exception):
+        entry["default"] = list(pt.defaultValue())
```
healkeiser: **entry["default"] = list(pt.defaultValue()) silently drops the default for Toggle, Menu and Ramp templates, because their defaultValue() returns a scalar, not a tuple.**

hou.ToggleParmTemplate.defaultValue() -> bool and hou.MenuParmTemplate.defaultValue() -> int (confirmed in $HFS/houdini/help/hom.zip). list(True) / list(2) raise TypeError, which contextlib.suppress swallows, so the entry carries no `default` key at all. Probed against the PR head with mocked hou: Toggle -> {'name','label','type'} only; Float -> {'default': [0.5]}. Every toggle and every menu in an interface loses its default while the tool advertises "every parameter in order with defaults". _template_to_dict, 870 lines above in the same file, already handles this with a nested fallback to the raw defaultValue().

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
 
 
 register_handler("parameters.get_parameters", _get_parameters)
+
+
+###### Handler: parameters.get_parm_references
+
+_CHANNEL_REF_RE = re.compile(r"""\bch[sfo]?\s*\(\s*['"]([^'"]+)['"]""")
```
healkeiser: **_CHANNEL_REF_RE matches only ch/chs/chf/cho, missing most real channel functions (and `cho` is not one of them), so outgoing silently under-reports references in non-pure expressions.**

$HFS/houdini/help/expressions.zip lists the channel-reading HScript functions as ch, chexist, chexpr, chexprf, chexprt, chf, chramp, chrampf, chrampraw, chrampt, chs, chsop, chsoplist, chsraw, cht. The regex covers 3 of 15 and adds a nonexistent `cho`. Probed: chramp("../src/r", 0.5) and chsop("../ctrl/path") both return []. A parm whose expression is `chramp("../ctrl/falloff", @u)` or `chsop("../ctrl/geo")` comes back with `expression` set, no `references` key and no `pure_reference` key, so the reply looks identical to a `$F`-only expression: a rename audit reports "nothing reads this" for a parm that is read.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+            entry["references"] = [direct.path()]
+            entry["pure_reference"] = True
+            return entry
+    references: list[str] = []
+    node = parm.node()
+    for token in _CHANNEL_REF_RE.findall(expression):
+        target = None
+        with contextlib.suppress(Exception):
+            target = node.parm(token)
+        references.append(target.path() if target is not None else token)
```
healkeiser: **references.append(target.path() if target is not None else token) mixes resolved absolute parm paths and unresolved raw expression tokens in one list with no marker distinguishing them.**

The PR's own test bakes this in: `references == ["/obj/src/scale", "../src/name"]`. A consumer reading the reply cannot tell whether "../src/name" is a reference that resolved to a relative path or a token that failed to resolve, and cannot feed the list to hou.parm() uniformly. Any node whose expression references a parm on a node that was deleted or renamed yields a path-shaped string that silently means "unknown".

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+    incoming: list[dict[str, Any]] = []
+    outgoing: list[dict[str, Any]] = []
+    truncated = False
+    for parm in parms:
+        if len(incoming) + len(outgoing) >= int(limit):
+            truncated = True
+            break
+        if direction in ("both", "incoming"):
+            with contextlib.suppress(Exception):
+                refs = [p.path() for p in parm.parmsReferencingThis()]
```
healkeiser, line 916: **parmsReferencingThis() is called once per parameter of the node and scans the whole scene each time; `limit` caps the reply, not the work.**

HOM documents parmsReferencingThis as returning "all the parameters in the scene that have channel references to this parameter". The default call (no parm_name, direction="both") on a Karma ROP or an HDA with a 500-instance multiparm iterates every hou.Parm from node.parms() and triggers a scene-wide dependency scan for each one. The `if len(incoming) + len(outgoing) >= limit: break` guard only stops once entries accumulate, so a node with 2000 parms and 2 references does 2000 scene scans before returning 2 rows, blocking Houdini's main thread.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+    node = _resolve_node(node_path)
+    if direction not in ("both", "incoming", "outgoing"):
+        raise ValueError("direction must be 'both', 'incoming' or 'outgoing'.")
+    parms = [_resolve_parm(node_path, parm_name)] if parm_name is not None else list(node.parms())
+
+    incoming: list[dict[str, Any]] = []
+    outgoing: list[dict[str, Any]] = []
+    truncated = False
+    for parm in parms:
+        if len(incoming) + len(outgoing) >= int(limit):
```
healkeiser: **`truncated: true` is reported whenever the entry count reaches `limit`, even when every remaining parameter would have produced nothing.**

Probed on the PR head: a node with two parms, the first with one incoming reference and the second with none, called with limit=1, returns truncated=True with the complete result. The loop breaks before inspecting the second parm and cannot know it was empty. A caller that retries with a higher limit, or that reports "results incomplete" to the user, is acting on a false signal.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+    result: dict[str, Any] = {
+        "node_path": node.path(),
+        "parm_name": parm_name,
+        "direction": direction,
+        "incoming": incoming,
+        "outgoing": outgoing,
+        "truncated": truncated,
+    }
+    with contextlib.suppress(Exception):
+        result["node_dependents"] = sorted(
```
healkeiser: **node_dependents / node_references use hou.OpNode.dependents()/references() with the default include_children=True, are unbounded by `limit`, and are read without cooking despite HOM recommending it.**

HOM (hou/OpNode.txt) documents both as recursing into children by default and adds "the result can differ depending last cook of the nodes. It's recommended that you first call cook() on the node first." Calling get_parm_references on a subnet or an HDA returns every node referenced by every descendant, ignoring the `limit` argument entirely, so the reply can be thousands of paths. On a freshly loaded, uncooked scene the same call can return an empty or stale list with no indication that the answer depends on cook state.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+            entry["pure_reference"] = True
+            return entry
+    references: list[str] = []
+    node = parm.node()
+    for token in _CHANNEL_REF_RE.findall(expression):
+        target = None
+        with contextlib.suppress(Exception):
+            target = node.parm(token)
+        references.append(target.path() if target is not None else token)
+    if references:
```
healkeiser: **_outgoing_reference returns an entry for any expression, including ones with no channel reference at all, so non-references consume the `limit` budget and pollute `outgoing`.**

Probed: a parm with expression `$F * 2` returns {'parm': 'tz', 'expression': '$F * 2'} with no `references` and no `pure_reference`. On an animated or procedurally-driven node, every `$F`, `$FF`, `rand()`, `fit()` and `point()` expression is emitted as an outgoing row. With limit=200 these can fill the budget before a real reference is reached, and the tool's contract ("what this node's expressions read, resolved to paths") is not what the reply contains.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+        if budget[0] <= 0:
+            break
+        budget[0] -= 1
+        if "children" in entry:
+            entry["children"] = _prune_tree(entry["children"], budget)
+        kept.append(entry)
+    return kept
+
+
+def _node_type_for_tree(context: str, type_name: str):
```
healkeiser, line 1051: **_node_type_for_tree re-implements graph_handlers._resolve_node_type verbatim, and re-implements the category lookup that node_handlers._list_node_types already does.**

graph_handlers.py:45 already has the identical preferredNodeType -> exact -> newest-versioned cascade with the same docstring rationale, and node_handlers.py:501 already has the hou.nodeTypeCategories().get(context) + sorted-available-categories error. graph_handlers already cross-imports from node_handlers (`from fxhoudinimcp_server.handlers.node_handlers import _resolve_input_index`), so the precedent for importing exists. This is now a third copy of the category lookup and a second copy of the version cascade; a fix to the version-resolution rule has to be made in two places or it drifts.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+
+register_handler("parameters.get_parm_references", _get_parm_references)
+
+
+###### Handler: parameters.get_parm_template_tree
+
+_TREE_MENU_CAP = 30
+
+
+def _template_tree_entry(pt: hou.ParmTemplate) -> dict[str, Any]:
```
healkeiser, line 965: **_template_tree_entry is a second, divergent serializer for hou.ParmTemplate alongside _template_to_dict in the same module.**

Both walk the same object and emit the same facts under different keys and different caps: default vs default_value, min_strict vs min_is_strict, hidden vs is_hidden, menu cap 30 (_TREE_MENU_CAP) vs hard-coded 50, components vs num_components. The older one handles the Toggle/Menu default correctly and the new one does not (see the defaultValue finding), which is exactly the cost of the duplication. The tree entry is a superset of the schema entry; one function with a `verbose`/`nested` flag would cover both and make the Toggle fix apply to get_parameter_schema too.

healkeiser (commented):
### Code review

15 findings. 14 are inline below. 1 land outside this diff and are listed here:

- `README.md`:12 - The tool-count bump was applied only to server_instructions.md; README.md and the docs still claim 199 tools and 12 parameter tools.

  Stale after this PR: README.md lines 12, 57, 102, 594, 607; docs/index.md lines 7, 52, 83; docs/how-to/index.md line 6; docs/how-to/tools.md line 5 ("199 tools") and line 56 ("### Parameters (12 tools)", now 14). Commit a73e32c shows the repo convention is to bump all of these together with the instructions header. Only server_instructions.md is guarded by a test (tests/test_prompts_resources.py::TestInstructionHeaderCounts), which is why the rest will drift unnoticed.


python/fxhoudinimcp/tools/parameters.py
```diff
@@ -1,6 +1,6 @@
 """MCP tools for Houdini parameter operations.
 
-Exposes 10 tools covering parameter get/set, expressions, channel
+Exposes 12 tools covering parameter get/set, expressions, channel
```
healkeiser: **The module docstring is changed to "Exposes 12 tools" but the file now defines 14 @mcp.tool() functions; the handler module docstring still says 10 of 14.**

Counted on the PR head: `grep -c '^@mcp.tool()' python/fxhoudinimcp/tools/parameters.py` = 14. The count was already stale at 10 (main has 12), and the PR edited the line without re-counting, so it stays wrong by the same 2. houdini/.../parameter_handlers.py line 3 ("Provides 10 command handlers") was not touched at all and is now off by 4.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+    """What *parm* reads from, or None when it holds a plain value.
+
+    A pure `ch("../src/tx")` resolves through getReferencedParm(). Anything
+    richer -- `ch("../src/scale") * 2` -- answers with the parm itself there
+    (measured on 22.0.429), so the channel references are read out of the
+    expression text and resolved relative to the node.
+    """
+    expression = None
+    with contextlib.suppress(Exception):
+        expression = parm.expression()
```
healkeiser, line 828: **String parameters carrying backtick channel references are invisible to `outgoing`, because parm.expression() does not return them.**

A cache path parm holding `$HIP/cache/`chs("../CTRL/version")`/geo.bgeo.sc` stores that as the raw string value, not an expression; parm.expression() raises hou.OperationFailed, the suppress swallows it and _outgoing_reference returns None. These backtick references are the most common cross-node link in a shot setup, and the rename-audit use case the tool is built for will miss all of them. hou.Parm.unexpandedString() is the accessor that sees them. Medium confidence: the code path was verified but the expression()/unexpandedString() split was not verified on a live 22.0 parm.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+            entry["tags"] = {
+                k: (v if len(str(v)) <= 120 else str(v)[:120] + "...") for k, v in tags.items()
+            }
+    if entry["type"] == "Folder":
+        with contextlib.suppress(Exception):
+            entry["folder_type"] = pt.folderType().name()
+        with contextlib.suppress(Exception):
+            if pt.endsTabGroup():
+                entry["ends_tab_group"] = True
+        with contextlib.suppress(Exception):
```
healkeiser, line 1005: **The whole children list comprehension sits inside one contextlib.suppress, so one bad child template drops every sibling and the folder reports as empty.**

_template_tree_entry's first three lines (pt.name(), pt.label(), _parm_type_name(pt)) are deliberately not suppressed, so a template whose label() or type() raises propagates out of the comprehension. The enclosing suppress at line 940 then discards the entire `children` list for that folder, and the folder is reported with no `children` key at all - indistinguishable from an empty tab. Wrapping each child instead of the list would lose one entry rather than the whole folder.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+            entry["children"] = [_template_tree_entry(child) for child in pt.parmTemplates()]
+        return entry
```
healkeiser, line 1011: **The Folder branch returns before reading defaultValue(), so a multiparm block's default instance count is never reported although the tool advertises multiparm blocks.**

hou.FolderParmTemplate.defaultValue() returns the default number of multiparm instances. `get_parm_template_tree` on an HDA with a MultiparmBlock folder shows folder_type="MultiparmBlock" and its child templates (with `#` in their names) but nothing about how many instances a fresh node gets, which is the one number a caller building that interface needs.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -795,3 +796,296 @@ def _get_parameters(
...
+) -> dict[str, Any]:
+    """Who references a parameter, and what it references -- both directions.
+
+    `incoming` lists, per parameter of *node_path* (or the one *parm_name*),
+    the parameters elsewhere whose expressions read it (parmsReferencingThis).
+    `outgoing` lists what this node's expressions read. Node-level
+    `dependents` / `references` round it off, so "what breaks if I rename
+    this control" is one call instead of a HOM script.
+    """
+    node = _resolve_node(node_path)
```
healkeiser: **`direction` is validated after _resolve_node, so a bad enum on a path that does not resolve reports the wrong error.**

get_parm_references("/obj/typo", direction="sideways") raises "Node not found: /obj/typo" and says nothing about the invalid direction; the caller fixes the path, calls again, and only then learns the second argument was also wrong. Cheap argument checks belong before the scene lookup.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Confirmed live (`True` / `1`). Fixed in 50b0725 by building tree entries on `_template_to_dict`, which already falls back to the scalar.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: all 15 channel functions from expressions.zip, `cho` removed; unit test on `chramp` / `chsop`.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: `references` holds resolved paths only, `unresolved` the written targets that no longer resolve (checked live with a `chs("../gone/x")`).

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725 by not scanning when nothing can be found: a reader through ch() or backticks makes its node a dependent of the target (measured; a self-reference makes the node its own), so a node with no dependents skips the per-parm scan. With dependents the scan is still per parameter — happy to go further (e.g. scan only the dependents' parms) if you see it bite.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: set only when an entry did not fit; unit test for both cases.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: `include_children=False`, capped at `limit` with a count past it, and a note when the node needs to cook. I did not cook from a read.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: `$F * 2` produces no entry (live test).

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: `_node_type_for_tree` calls `graph_handlers._resolve_node_type`; only the category lookup and its error stay here.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: a tree entry is `_template_to_dict(pt)` plus tree-only keys, so the keys match get_parameter_schema (`default_value`, `is_hidden`, `min_is_strict`, `menu_items`) and fixes land in both.

python/fxhoudinimcp/tools/parameters.py
d-potrivaev: Fixed in 50b0725: 14 and 14.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Confirmed live (the medium-confidence part too: `expression()` raises OperationFailed, `unexpandedString()` has the text). Fixed in 50b0725: backtick segments are read from `unexpandedString()` and the entry says `in_backticks`; live test sees the reference from both ends.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: each child is guarded on its own.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725: `default_instances` (live test: 3).

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 50b0725.

d-potrivaev:
Thanks for the review. All 15 findings are answered in 50b0725, one commit on top of the branch: 15 addressed as suggested. Each inline thread has its answer.

Findings outside the diff:
- **Tool count in README/docs.** Fixed in 50b0725: 201 tools, Parameters 14.

Verified: ruff check + format, `pytest -m "not integration"` (591 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_parameters_live.py::TestParmReferences / TestParmTemplateTree) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: findings are fixed or correctly rebutted, unit tests pass (591). The cook-state note wording (incoming now depends on dependents()) will be adjusted on main after merge.

PR #79, opened 2026-09-16
fix(parameters): get_parameter describes a Data parameter instead of answering null

d-potrivaev:
## What this changes

A Data parameter -- a Curve SOP's stash, a Stash SOP's geometry -- holds a
blob: eval() is a hou.Geometry or None and rawValue() is empty either way
(measured on 22.0.429). get_parameter therefore answered `value: null`
whether the parameter was empty or held a whole mesh, and a caller had no
way to tell the two apart short of execute_python.

Both readers, get_parameter and get_parameters, now go through one helper: `value` stays
the geometry summary a set blob always serialised to (None when unset), and `data:
{is_set, data_parm_type, geometry}` is added next to it, with the counts from
serialize.geometry_summary. Live: a Stash SOP before Stash Input reports is_set false;
after it, 8 points, 6 primitives, 24 vertices. Other parameter types are untouched.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 575 passed (`tests/test_data_parm.py` and the updated neighbours)
- [ ] Generated files regenerated — none touched

Updated after review in 81970c2.


houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -73,6 +73,32 @@ def _serialize_value(value: Any) -> Any:
...
+        if isinstance(value, hou.Geometry):
+            with contextlib.suppress(Exception):
+                summary["geometry"] = {
+                    "points": value.intrinsicValue("pointcount"),
+                    "prims": value.intrinsicValue("primitivecount"),
+                }
+        else:
+            summary["value_type"] = type(value).__name__
+    with contextlib.suppress(Exception):
+        blob = parm.asData()
```
healkeiser: **hou.Parm has no asData() method, so the entire size_bytes / is_set-rescue block is dead code silently swallowed by contextlib.suppress.**

Verified against the shipped HOM stub: `grep -c asData hou.py` returns 0, and enumerating class Parm's 143 methods shows eval/evalAsGeometry/evalAsJSONMap but no asData. So parm.asData() raises AttributeError, contextlib.suppress(Exception) eats it, and get_parameter on a Stash SOP never emits size_bytes and never sets is_set from the blob. The PR body's live claim ('1275 bytes') cannot come from this code path. The intended API is parm.evalAsJSONMap() (dataParmType is only Geometry or KeyValueDictionary).

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -73,6 +73,32 @@ def _serialize_value(value: Any) -> Any:
...
 
+def _data_parm_summary(parm: hou.Parm, pt: hou.ParmTemplate) -> dict[str, Any]:
+    """Size and content of a Data parameter, never the blob itself."""
+    summary: dict[str, Any] = {"is_set": False}
+    with contextlib.suppress(Exception):
+        summary["data_parm_type"] = pt.dataParmType().name()
+    value = None
+    with contextlib.suppress(Exception):
+        value = parm.eval()
+    if value is not None:
```
healkeiser: **`if value is not None` reports is_set=True for an empty KeyValueDictionary Data parameter, the exact false positive the PR set out to remove.**

A Data parm with dataParmType KeyValueDictionary evals to `{}` when empty, not None, so is_set becomes True and value_type becomes 'dict' for a parameter holding nothing. Same for a hou.Geometry object with 0 points. With the asData() finding killing the size_bytes correction, nothing downstream can fix it: the caller is told the blob is set when it is empty.

healkeiser (commented):
### Code review

11 findings. 10 are inline below. 1 land outside this diff and are listed here:

- `houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py`:819 - The sibling batch reader parameters.get_parameters still answers a Data parameter the old ambiguous way, so the root cause is only patched on the single-parm path.

  `get_parameters(node_path="/obj/geo1/stash1")` runs `_serialize_value(parm.eval())` on the `stash` parm: unset -> `value: null`, set -> the raw geometry summary, with no is_set and no data_parm_type. discipline.md explicitly steers the model to get_parameters ('reads many live values at once'), so the recommended path keeps the bug. The fix belongs in a shared helper both readers route through, not in an early-return branch of one handler.


houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -73,6 +73,32 @@ def _serialize_value(value: Any) -> Any:
...
+        summary["is_set"] = True
+        if isinstance(value, hou.Geometry):
+            with contextlib.suppress(Exception):
+                summary["geometry"] = {
```
healkeiser: **_data_parm_summary re-implements serialize.geometry_summary with incompatible key names and drops vertex_count.**

houdini/scripts/python/fxhoudinimcp_server/serialize.py:82 already returns {"type":"Geometry","point_count":N,"prim_count":N,"vertex_count":N} from the same intrinsics, and every other handler's geometry answer goes through it (tests/test_hom_serialize.py:227 pins that shape). get_parameter now emits a second, conflicting shape {"points":N,"prims":N}, so a client parsing geometry counts needs two code paths and loses vertex_count. Call serialize.geometry_summary instead.

tests/test_data_parm.py
```diff
@@ -0,0 +1,90 @@
...
+class TestDataParmSummary:
+    def test_a_set_geometry_blob_reports_its_counts(self):
+        geometry = _geometry_type()
+        geometry.intrinsicValue.side_effect = lambda name: {
+            "pointcount": 8,
+            "primitivecount": 6,
+        }[name]
+        parm = MagicMock()
+        parm.eval.return_value = geometry
+        parm.asData.return_value = {"geometry": "x" * 120}
```
healkeiser: **The new tests use a bare MagicMock for parm, so parm.asData() auto-succeeds and the tests cannot fail on the nonexistent HOM method they assert against.**

`parm = MagicMock()` with no spec auto-creates `asData`, so `assert summary["size_bytes"] == 120` and `== 0` pass green while production never produces the key at all. Spec'ing the mock to hou.Parm's real method list, or adding a case to tests/integration/test_parameters_live.py, is what would have caught it; the PR ran neither.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -131,6 +157,23 @@ def _get_parameter(node_path: str, parm_name: str, **_: Any) -> dict[str, Any]:
     parm = _resolve_parm(node_path, parm_name)
     pt = parm.parmTemplate()
 
+    if _parm_type_name(pt) == "Data":
+        # A Data parameter (a Curve SOP's stash, a Stash SOP's geometry) holds
+        # a blob: eval() is a hou.Geometry or None and rawValue() is empty
+        # either way, so the old answer was `null` with no way to tell "unset"
+        # from "set". Report the size and what it holds instead.
+        return {
```
healkeiser: **The Data branch drops the `value` and `raw_value` keys the handler has always returned, breaking any caller that indexes them.**

Before this PR a set Data parm returned `value` as the standard geometry summary (hou.Geometry hits json_default -> serialize.geometry_summary in hwebserver_app._json_response), not null as the PR body states; only the unset case was null. Any client or prompt doing `result["value"]` on a Data parm now raises KeyError instead of getting a summary, and the field simply vanishes with no deprecation.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -73,6 +73,32 @@ def _serialize_value(value: Any) -> Any:
...
+                summary["geometry"] = {
+                    "points": value.intrinsicValue("pointcount"),
+                    "prims": value.intrinsicValue("primitivecount"),
+                }
+        else:
+            summary["value_type"] = type(value).__name__
+    with contextlib.suppress(Exception):
+        blob = parm.asData()
+        if isinstance(blob, dict):
+            summary["size_bytes"] = sum(len(str(v)) for v in blob.values())
```
healkeiser: **`sum(len(str(v)) for v in blob.values())` is not a byte count, so the field name size_bytes lies.**

If the call is corrected to evalAsJSONMap(), values are str and len() is a character count, not bytes (any non-ASCII inflates or deflates it). If any value is bytes, `str(b'\x00\x01')` yields "b'\\x00\\x01'" -> 12 for 2 bytes, roughly 4x over-reporting, which also flips `is_set` true for an all-empty-bytes blob. Use `len(v.encode())` / `len(v)` on the actual payload, or rename the field.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
```diff
@@ -73,6 +73,32 @@ def _serialize_value(value: Any) -> Any:
...
+    summary: dict[str, Any] = {"is_set": False}
+    with contextlib.suppress(Exception):
+        summary["data_parm_type"] = pt.dataParmType().name()
+    value = None
+    with contextlib.suppress(Exception):
+        value = parm.eval()
+    if value is not None:
+        summary["is_set"] = True
+        if isinstance(value, hou.Geometry):
+            with contextlib.suppress(Exception):
```
healkeiser: **One contextlib.suppress wraps both intrinsicValue calls, so a failure on either drops the whole geometry dict instead of the one count.**

The dict literal is built before assignment, so if intrinsicValue('primitivecount') raises on some geometry type, the successful pointcount is discarded too and `geometry` is absent entirely, with no indication why. serialize.geometry_summary suppresses per-key for exactly this reason.

tests/test_data_parm.py
```diff
@@ -0,0 +1,90 @@
...
+sys.modules.setdefault("hou", MagicMock())
+sys.modules.setdefault("hdefereval", MagicMock())
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "houdini", "scripts", "python"))
+
+# Internal
+import fxhoudinimcp_server.handlers.parameter_handlers as parameters  # noqa: E402
+
+
+def _geometry_type():
+    geometry = MagicMock(spec=["intrinsicValue"])
```
healkeiser: **_geometry_type() mutates the process-wide sys.modules['hou'] MagicMock and never restores it.**

`parameters.hou.Geometry = type(geometry)` writes to the MagicMock shared by ~15 other test modules that do sys.modules.setdefault('hou', MagicMock()), leaving hou.Geometry a real type for the rest of the session. tests/test_hom_serialize.py:140 shows the repo's own save/swap/restore pattern; without it, any future isinstance(x, hou.Geometry) check in another unit test becomes order-dependent.

tests/test_data_parm.py
```diff
@@ -0,0 +1,90 @@
...
+        template = MagicMock()
+        template.dataParmType.return_value.name.return_value = "Geometry"
+        summary = parameters._data_parm_summary(parm, template)
+        assert summary["is_set"] is False
+        assert summary["size_bytes"] == 0
+
+
+class TestGetParameterOnData:
+    def test_a_data_parameter_answers_with_the_summary_not_a_value(self, monkeypatch):
+        _geometry_type()
```
healkeiser: **TestGetParameterOnData calls _geometry_type() and discards the mock, so the isinstance(hou.Geometry) branch is never exercised there.**

parm.eval() returns None in that test, so the geometry branch is unreachable and the _geometry_type() call is pure ceremony that only performs the global mutation described in the previous finding. Dropping the call leaves the test identical.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
healkeiser: **The early-return duplicates nine keys of the result dict, so any field later added to the ordinary path silently goes missing on Data parameters.**

node_path, parm_name, parm_type, is_locked, is_at_default, expression, expression_language and keyframe_count are spelled twice (and parm_type is hardcoded to the literal 'Data' rather than _parm_type_name(pt), while parm_name comes from parm.name() here and from the argument below). Building the shared dict once and swapping value/raw_value for data on the Data path is smaller and cannot drift.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: `hou.Parm.asData()` does exist on 22.0 (`hou/Parm.txt`, "asData(value=True, evaluate_value=False, ...)"); the 1275 came from it, measured live on a Stash SOP. The finding still led somewhere right: on a Data parm it returns `{"geometry": <the whole geometry serialised to a string>}`, so the number was a character count and producing it serialised the blob on the main thread. 81970c2 drops `size_bytes` and the `asData()` call; `evalAsJSONMap()` is not the alternative either — on this parm it raises "Cannot evaluate this parameter as a JSON map".

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 81970c2: a dict value is set only when non-empty (`key_count` reported). A geometry is set even when empty, and its counts say so.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 81970c2: `serialize.geometry_summary`, so `point_count` / `prim_count` / `vertex_count` as everywhere else.

tests/test_data_parm.py
d-potrivaev: The method exists, but the point about unspecced mocks holds: 81970c2 specs the parm mock and adds a live test (`TestDataParameter` in `test_parameters_live.py`) that reads a Stash SOP before and after Stash Input through both readers.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Right — a set blob did serialise to the geometry summary. Fixed in 81970c2: `value` and `raw_value` are back with their old meaning; `data` is added next to them.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Removed in 81970c2 (see the `asData` thread).

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 81970c2 by using `serialize.geometry_summary`, which suppresses per key (unit test drops one intrinsic).

tests/test_data_parm.py
d-potrivaev: Fixed in 81970c2: `hou.Geometry`, `hou.OperationFailed` and friends are set through monkeypatch only.

tests/test_data_parm.py
d-potrivaev: Gone in 81970c2.

houdini/scripts/python/fxhoudinimcp_server/handlers/parameter_handlers.py
d-potrivaev: Fixed in 81970c2: no early return; the one result dict gets `value` from the Data helper and `data` added.

d-potrivaev:
Thanks for the review. All 11 findings are answered in 81970c2, one commit on top of the branch: 10 addressed as suggested, 1 not reproduced on 22.0.429 (measurements in the thread). Each inline thread has its answer.

Findings outside the diff:
- **get_parameters (parameter_handlers.py:819).** Fixed in 81970c2: both readers go through `_data_parm_value`; checked live through `get_parameters(patterns=["stash"])`.

Verified: ruff check + format, `pytest -m "not integration"` (575 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_parameters_live.py::TestDataParameter) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head. You were right on hou.Parm.asData(): it lives in houpythonportion/Parm.py and is grafted onto hou.Parm, my grep of hou.py was inconclusive. Unit tests pass (575). One leftover, a set KeyValueDictionary parm now returns value: None, will be fixed on main after merge.

PR #80, opened 2026-09-16
feat(nodes): change_node_type swaps a node's type in place, and press_button presses a button parameter and reads the node back

d-potrivaev:
## What this changes

Two Houdini gestures had no tool. Moving an asset instance to an installed
newer version (Type Properties "change type"), or swapping a placeholder for
the real node, meant hou.Node.changeNodeType from execute_python -- and a
call written without HOM's defaults in mind dropped parameter values with
nothing in the reply to say so. Pressing a button parameter ("Stash Input",
"Reload Geometry", an asset's own Build button) was execute_python too, and
whatever the node had to say afterwards stayed in Houdini.

change_node_type resolves the new type in the node's own category the way
create_node does (preferred version for an unversioned name), keeps the
name, wires, position, flags, parameter values and network contents by
default, and reports `parms_dropped`: the values that had been set and have
no home on the new type, the one thing the swap can lose quietly. The same
type is a no-op with `changed: false`. Measured on Houdini 22.0.429: an
xform with tx=2.5 changed to a null keeps its wire from the box upstream
and names `tx` as dropped.

press_button runs the callback as a click would, times it, and returns the
node's errors and warnings afterwards; a wrong name gets a did-you-mean and
the list of the node's buttons; a non-button parameter is pressed but
noted. A Stash SOP's "Stash Input" took 0.2 ms and left 8 points in the
stash.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 585 passed (`tests/test_node_verbs.py` and the updated neighbours)

Updated after review in bb06d36.


houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+    with contextlib.suppress(Exception):
+        if changed.type().definition() is not None:
+            result["matches_definition"] = changed.matchesCurrentDefinition()
+    return result
+
+
+###### nodes.press_button
+
+
+def press_button(node_path: str, parm_name: str, arguments: dict | None = None) -> dict:
```
healkeiser: **press_button runs an arbitrary, potentially long callback under the default deadline (60 s client, 120 s server) instead of being registered as an unbounded command, so exactly the presses it advertises time out while Houdini keeps working.**

press_button(node_path='/obj/geo1/filecache1', parm_name='execute') on a heavy cache: httpx (bridge.py:81, timeout=60.0) aborts at 60 s and the agent is told the call failed, while Houdini is still writing. dispatcher.py:46 lists only cache.write_cache and rendering.start_render in _UNBOUNDED_COMMANDS, and _TIMEOUT_HINTS['code.execute_python'] (dispatcher.py:61-64) explicitly warns 'If this was a cook, a render or a Save to Disk, use write_cache / start_render instead of pressing buttons in Python' - the new tool reintroduces that footgun as a first-class command with no hint and no exemption. The docstring's claim 'The call holds until the callback returns' is false past 60 s.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+            f"Callback of {node.path()}/{parm_name} failed: {readable_message(exc)}"
+        ) from exc
+    duration_ms = round((time.perf_counter() - started) * 1000, 1)
+    result: dict[str, Any] = {
+        "success": True,
+        "node_path": node.path(),
+        "parm_name": parm.name(),
+        "parm_type": parm_type,
+        "duration_ms": duration_ms,
+        "errors": list(node.errors()),
```
healkeiser: **errors/warnings are read straight after pressButton, but HOM defines them as the messages from the node's LAST COOK - pressing a button usually only dirties the node, so the fields describe a state that predates the press.**

A button whose callback puts the node into an error state (bad reload path, failed asset Build) returns {'success': True, 'errors': []} because nothing cooked between the press and the read. Conversely a node that errored on a previous cook reports those stale errors as if the press caused them. hou/Node.txt: 'Return the text of any errors from the last cook of this node'. The tool's stated purpose - 'read what the node says afterwards' - is exactly what it fails to deliver; it needs a cook (or to say the values are pre-press).

healkeiser: **node.errors() and node.warnings() are called unguarded while get_node_info in this same file wraps both in try/except, so a raise here throws away the result of a button press that already happened.**

Same file, lines 355-362, defensively does `try: errors = list(node.errors()) except Exception: errors = []`. press_button omits that guard. On a node type where errors() raises, the handler raises after parm.pressButton() has already run: the side effect is committed, the caller gets an exception, and cannot tell whether the button fired. Every other field in the dict (callback_present, child_count, matches_definition) is suppressed with contextlib.suppress; these two are not.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
     }
 
 
+###### nodes.change_node_type
+
+
+def _resolve_node_type(category: hou.NodeTypeCategory, type_name: str):
```
healkeiser: **_resolve_node_type is a verbatim copy of the identical helper already in graph_handlers.py:45, including the same fallback chain and the same latent sort bug.**

graph_handlers.py:45-62 has byte-for-byte the same body (preferredNodeType -> exact key -> newest versioned -> None), used by build_network and get_node_card. Two copies now drift independently: any fix to version resolution has to be found and applied twice, and a reader grepping `_resolve_node_type` gets two different modules. Move it to a shared module (e.g. alongside _get_node) and import it in both.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+    newest versioned name."""
+    with contextlib.suppress(Exception):
+        preferred = hou.preferredNodeType(f"{category.name()}/{type_name}")
+        if preferred is not None:
+            return preferred
+    types = category.nodeTypes()
+    if type_name in types:
+        return types[type_name]
+    prefix = type_name + "::"
+    versioned = sorted(key for key in types if key.startswith(prefix))
```
healkeiser: **The 'newest versioned name' fallback sorts version keys lexicographically, so a two-digit version loses to a one-digit one.**

sorted(['mytool::10.0', 'mytool::2.0'])[-1] == 'mytool::2.0'. change_node_type('/obj/geo1/n', 'mytool') on a scene where hou.preferredNodeType is unavailable or returns None (the suppressed-exception path at line 603) silently downgrades the node to version 2.0 instead of upgrading it to 10.0 - the exact opposite of the 'upgrade to version' gesture the docstring advertises. Needs a version-aware key, not string sort.

healkeiser (commented):
### Code review

15 findings. 13 are inline below. 2 land outside this diff and are listed here:

- `tests/integration/test_path_validation_live.py`:78 - change_node_type adds a required argument `new_type` that is not in STAND_INS, so the self-discovering path-validation test fails on the new command.

  test_path_validation_live._cases() walks the handler registry; nodes.change_node_type has node_path (a PATH_PARAM) plus required new_type with no default and no STAND_INS entry, so the case becomes {'__missing__': ['new_type']} and test_unresolvable_path_is_rejected_and_named[nodes.change_node_type] fails with "has required arguments this test does not know how to fill: ['new_type']. Add them to STAND_INS." Verified by inspecting the registered signature: [('node_path', required), ('new_type', required), ('keep_name', default), ...]. press_button is fine because parm_name is already in STAND_INS. Fix: add "new_type": "null" to STAND_INS.

- `docs/how-to/tools.md`:5 - The tool count was bumped from 199 to 201 only in server_instructions.md, leaving README.md and the docs site claiming 199.

  tests/test_prompts_resources.py::TestInstructionHeaderCounts only checks server_instructions.md, so CI passes while docs/how-to/tools.md:5, docs/how-to/index.md:6, docs/index.md:7/52/83, README.md:12/57/102/594/607 still say 199 tools - and docs/how-to/tools.md:49 still says 'Node Operations (20 tools)' for what is now 22. README.md:594 additionally claims 'tests/integration/ executes all 199 commands', which is now wrong in two ways since neither new command has an integration test.


houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+    """
+    node = _get_node(node_path)
+    category = node.type().category()
+    resolved = _resolve_node_type(category, new_type)
+    if resolved is None:
+        close = get_close_matches(new_type, list(category.nodeTypes()), n=3, cutoff=0.5)
+        hint = f" Did you mean: {close}?" if close else ""
+        raise ValueError(f"Type '{new_type}' does not exist in {category.name()}.{hint}")
+    old_type = node.type().name()
+    if resolved.name() == old_type:
```
healkeiser: **The same-type early return makes the documented `keep_network_contents=False` reset unreachable, because HOM exposes it through force_change_on_node_type_match which this handler never passes.**

The docstring says keep_network_contents=False 'resets an asset to its definition's contents'. The commonest reason to do that is on an asset already of the target type (its definition changed on disk, the instance is dirty). change_node_type('/obj/myasset', 'myasset::2.0', keep_network_contents=False) hits `if resolved.name() == old_type` and returns {'changed': False, 'message': '...nothing changed.'} without touching the node. HOM hou/OpNode.txt documents changeNodeType(..., force_change_on_node_type_match=False) as 'whether to perform the change even when is already of the specified type' - the parameter that would make this work is neither exposed nor passed.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+        "success": True,
+        "node_path": changed.path(),
+        "old_type": old_type,
+        "new_type": changed.type().name(),
+        "changed": True,
+        "keep_parms": bool(keep_parms),
+        "keep_network_contents": bool(keep_network_contents),
+        # Values that had been set and have no home on the new type: the
+        # only thing the swap can lose without saying so.
+        "parms_dropped": sorted(before_non_default - after_parms),
```
healkeiser: **With keep_parms=False every non-default value is discarded, but parms_dropped computes a name-set difference and therefore reports an empty list.**

change_node_type('/obj/geo1/xform1', 'xform', keep_parms=False) - or any swap between two types sharing parm names - wipes tx/ty/tz, yet before_non_default - after_parms is empty because the names still exist on the new type. Result: {'success': True, 'keep_parms': False, 'parms_dropped': []}. The inline comment at line 684 claims this field is 'the only thing the swap can lose without saying so', so a caller trusting it concludes nothing was lost. Either refuse to compute parms_dropped when keep_parms is False, or report the reset values explicitly.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+            "success": True,
+            "node_path": node.path(),
+            "old_type": old_type,
+            "new_type": old_type,
+            "changed": False,
+            "message": f"{node.path()} is already of type {old_type}; nothing changed.",
+        }
+
+    before_parms = {p.name() for p in node.parms()}
+    before_non_default = {p.name() for p in node.parms() if not p.isAtDefault() and not p.isSpare()}
```
healkeiser: **before_non_default filters out spare parameters, so spare parms destroyed by the type swap never appear in parms_dropped - while parms_dropped_count does count them.**

`not p.isSpare()` excludes exactly the parameters the MCP's own workflow creates (server_instructions rule 6: 'Tweakables on a CTRL null with spare parms', plus the create_spare_parameter/create_spare_parameters tools). Swap a node carrying spare tweakables and every one of them is gone from the node with parms_dropped == [] and no other signal. Spare parms with values set are the highest-value entry for a field whose stated job is to name what the swap loses.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+        "node_path": changed.path(),
+        "old_type": old_type,
+        "new_type": changed.type().name(),
+        "changed": True,
+        "keep_parms": bool(keep_parms),
+        "keep_network_contents": bool(keep_network_contents),
+        # Values that had been set and have no home on the new type: the
+        # only thing the swap can lose without saying so.
+        "parms_dropped": sorted(before_non_default - after_parms),
+        "parms_dropped_count": len(before_parms - after_parms),
```
healkeiser: **parms_dropped_count counts a different set than parms_dropped lists, so it is never the length of the list it appears to count.**

parms_dropped = sorted(before_non_default - after_parms) (non-default, non-spare) but parms_dropped_count = len(before_parms - after_parms) (every parm, defaults included). Swapping a box SOP for a null yields something like {'parms_dropped': [], 'parms_dropped_count': 14}, and an agent reads '14 parameters dropped' as 14 lost values when none were set. Either name it parms_removed_count and say so, or make the two agree.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+    node = _get_node(node_path)
+    category = node.type().category()
+    resolved = _resolve_node_type(category, new_type)
+    if resolved is None:
+        close = get_close_matches(new_type, list(category.nodeTypes()), n=3, cutoff=0.5)
+        hint = f" Did you mean: {close}?" if close else ""
+        raise ValueError(f"Type '{new_type}' does not exist in {category.name()}.{hint}")
+    old_type = node.type().name()
+    if resolved.name() == old_type:
+        return {
```
healkeiser: **The no-op branch returns a dict missing every key the success branch provides, so a caller that reads the documented fields gets a KeyError.**

When resolved.name() == old_type the returned dict has only success/node_path/old_type/new_type/changed/message. Any caller (or agent-written follow-up) doing result['parms_dropped'], result['inputs'], result['outputs'], result['keep_parms'] or result['child_count'] - all present on every other return - raises KeyError. Conversely the success branch has no 'message' key. Two return shapes from one command with no way to tell them apart except branching on `changed`.

tests/test_node_verbs.py
```diff
@@ -0,0 +1,152 @@
...
+        category = MagicMock()
+        category.name.return_value = "Sop"
+        category.nodeTypes.return_value = {"curve": "classic", "curve::2.0": "new"}
+        hou.preferredNodeType.return_value = None
+        assert nodes._resolve_node_type(category, "curve") == "classic"
+        assert nodes._resolve_node_type(category, "curve::2.0") == "new"
+        category.nodeTypes.return_value = {"copytopoints::2.0": "v2", "copytopoints::3.0": "v3"}
+        assert nodes._resolve_node_type(category, "copytopoints") == "v3"
+        assert nodes._resolve_node_type(category, "nope") is None
+        hou.preferredNodeType.return_value = "preferred"
```
healkeiser: **TestResolveNodeType mutates the session-wide `hou` MagicMock without monkeypatch and never restores it, leaking hou.preferredNodeType.return_value = 'preferred' into every test module that runs afterwards.**

Demonstrated: appending a probe module that calls graph_handlers._resolve_node_type after tests/test_node_verbs.py returns the string 'preferred' instead of resolving through category.nodeTypes(). tests/test_node_versions.py, test_signature_menu.py, test_storm_session_fixes.py and test_update_mode_warning.py all sort after test_node_verbs.py and touch graph_handlers/build_network; today none happens to hit that path, so the suite passes by collection order alone. Every other test in this repo uses monkeypatch.setattr (test_config.py:83, test_material_assignments.py:151, test_node_placement.py:165) precisely because the mock is shared via sys.modules.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+        "success": True,
+        "node_path": node.path(),
+        "parm_name": parm.name(),
+        "parm_type": parm_type,
+        "duration_ms": duration_ms,
+        "errors": list(node.errors()),
+        "warnings": list(node.warnings()),
+    }
+    with contextlib.suppress(Exception):
+        callback = template.scriptCallback()
```
healkeiser: **callback_present is derived from ParmTemplate.scriptCallback(), which is empty for buttons implemented natively, so the tool reports False for the very examples its docstring names.**

'Reload Geometry' on a File SOP and 'Stash Input' on a Stash SOP are built-in (C++) button handlers with no script callback, so scriptCallback() returns '' and the result says callback_present: False after a press that did real work. An agent reading that field concludes the button is inert and either retries or falls back to execute_python. Either drop the field or rename it to say it only reports a *script* callback.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
```diff
@@ -592,6 +593,168 @@ def connect_nodes(
...
+        )
+    template = parm.parmTemplate()
+    parm_type = template.type().name()
+    started = time.perf_counter()
+    try:
+        if arguments:
+            parm.pressButton(dict(arguments))
+        else:
+            parm.pressButton()
+    except Exception as exc:
```
healkeiser, line 755: **A TypeError raised by pressButton for an unsupported argument value type is relabelled as a callback failure, blaming the node's script for a caller-side mistake.**

HOM hou/Parm.txt: pressButton 'Raises TypeError if an argument value type is unsupported' - values must be int, bool, float or str. The MCP schema is dict[str, Any], so a client can send arguments={'items': [1, 2]}; the handler reports "Callback of /obj/geo1/n/build failed: ..." although the callback never ran. The repo's own contract (tests/integration/test_bad_values_live.py::test_a_wrong_type_names_the_argument_and_its_type) requires a rejected value to name the argument and its type. Validate the argument values before pressing.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: `nodes.press_button` is in `_UNBOUNDED_COMMANDS` with a timeout hint, and the client passes `NO_TIMEOUT`, as for write_cache / start_render.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Confirmed live on 22.0.429: after pressing File's Reload on a bad path `errors()` is empty and `needsToCook()` True until a cook. Fixed in bb06d36: `cook=True` cooks after the press; otherwise `needs_cook` says the messages predate it.

d-potrivaev: Fixed in bb06d36: both reads are guarded; a press that happened is never lost to them.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: the copy is gone; change_node_type imports graph_handlers' one (function-local, since #76 makes graph_handlers import node_handlers).

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36 in the shared helper: `max` by `hda_handlers._version_key`, so `::10.0` beats `::2.0`.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: same type + `keep_network_contents=False` calls `changeNodeType(..., force_change_on_node_type_match=True)`.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Measured on 22.0.429: `keep_parms=False` did not reset `tx` on box→tube nor on xform→xform (forced), so the scenario does not occur there — but the point stands that the reply should not infer loss from the flags. bb06d36 measures it on the node: `parms_dropped` (set, no home) and `parms_reset` (set, back at default).

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Measured on 22.0.429: spare parms survive `changeNodeType` with their values (a `tweak` float stayed 3.0 on null→xform; live test `test_spare_parms_survive_the_swap`). The `isSpare()` filter is gone anyway, so a spare parm that did get lost would be reported.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: renamed `parms_removed_count` and documented as every parm the new type lacks, set or not.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: one reply shape; the no-op adds `message`.

tests/test_node_verbs.py
d-potrivaev: Fixed in bb06d36: `monkeypatch.setattr` on `preferredNodeType`.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Confirmed live: `scriptCallback()` is `''` for File's `reload` and Stash's `stashinput`. Renamed `has_script_callback`, docstring says built-in buttons have none and still do work.

houdini/scripts/python/fxhoudinimcp_server/handlers/node_handlers.py
d-potrivaev: Fixed in bb06d36: values are checked (int, bool, float, str) before pressing; the refusal names the key and type and says nothing was pressed.

d-potrivaev:
Thanks for the review. All 15 findings are answered in bb06d36, one commit on top of the branch: 13 addressed as suggested, 1 addressed differently or in part, 1 not reproduced on 22.0.429 (measurements in the thread). Each inline thread has its answer.

Findings outside the diff:
- **STAND_INS (test_path_validation_live.py:78).** Fixed in bb06d36: `"new_type": "null"`; both commands pass `test_unresolvable_path_is_rejected_and_named` live.
- **Tool count in README/docs.** Fixed in bb06d36: README and docs say 201, Node Operations 22; live tests for both commands added to `test_nodes_live.py`.

Verified: ruff check + format, `pytest -m "not integration"` (585 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_nodes_live.py::TestChangeNodeType / TestPressButton) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: all fixed, unit tests pass. Merging.

PR #81, opened 2026-09-16
refactor(scene): load_scene merges named nodes from another hip file and reports what arrived

d-potrivaev:
## What this changes

load_scene(merge=True) pulls a whole file into the scene and says nothing
about what came in. Bringing over a few named nodes -- a building generator
from the previous shot, one material network -- meant hou.hipFile.merge from
execute_python, and its node_pattern has two surprises, measured on Houdini
22.0.429: a relative name (`null1`) matches nothing, and a container named
alone (`/obj/geo1`) arrives without its children while `/obj/geo1/*` on its
own raises "Missing the parent of a node on load".

load_scene(merge=True) now takes node_paths (absolute; a relative one is refused with
the absolute form suggested, a pattern is refused) and overwrite_on_conflict. Each path
is sent as `<path> <path>/*`, so a container brings its children, and the reply is
attributed per requested path: `merged_nodes` (with `merged_descendant_count`),
`conflicts` from hou.hipFile.collisionNodesIfMerged (renumbered `null1` -> `null2`, or
overwritten in place) and `not_found_in_file`. A node merged into an existing container
(`/obj/geo1/sphere1`) is found. Load warnings come back as data for load and merge alike,
and a `$HIP/...` path is expanded before the existence check. No new tool: the first cut
added merge_hip, and review asked for parameters on load_scene instead.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 581 passed (`tests/test_load_scene_merge.py` and the updated neighbours)
- [ ] Generated files regenerated — none touched (no new command)

Updated after review in 8658b34.


houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+        node_paths: Absolute node paths to merge; default everything.
+        overwrite_on_conflict: Overwrite same-named nodes instead of
+            renaming the merged copy.
+    """
+    require_inside_project_root(file_path, "hip file")
+    if not os.path.isfile(file_path):
+        raise FileNotFoundError(f"File not found: {file_path}")
+    if node_paths is not None and not isinstance(node_paths, (list, tuple)):
+        raise ValueError("node_paths must be a list of absolute node paths.")
+    requested = [str(p).rstrip("/") for p in (node_paths or [])]
```
healkeiser, line 178: **An empty node_paths list falls through `node_paths or []` and becomes the `"*"` pattern, so "merge nothing" merges the entire hip file.**

An agent computes the nodes to bring over, the list comes back empty, and it calls merge_hip(path, node_paths=[]). The tool wrapper only skips the key when node_paths is None, so [] reaches the handler; requested == [] so pattern == "*" and every node in the other hip is dumped into the live scene. Verified with the mocked hou: pattern `*`, merged ['/obj/a','/obj/b','/obj/c']. Distinguish `None` from `[]` (empty list -> return merged_nodes: [] without calling merge).

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+    merged.sort()
+
+    conflicts: list[dict] = []
+    for path in existed:
+        entry = {"requested": path, "existed": True}
+        if overwrite_on_conflict:
+            entry["outcome"] = "overwritten in place"
+        else:
+            parent, _, name = path.rpartition("/")
+            renamed = [
```
healkeiser: **The `renamed` heuristic claims every newly merged sibling under the same parent as the rename of the colliding node, so unrelated nodes are reported as `merged_as`.**

merge_hip(f, ['/obj/null1', '/obj/building']) where /obj/null1 already exists. merged == ['/obj/building', '/obj/null2']; the filter keeps any merged path whose parent is /obj and whose basename != 'null1', so conflicts[0]['merged_as'] == ['/obj/building', '/obj/null2'] - /obj/building is a genuinely new node, not a rename. Confirmed by running the handler against the mocked hou. Restrict the candidates to paths not attributable to another requested path, or match Houdini's numeric-suffix rename (null1 -> null<N>).

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+                p for p in merged if p.rpartition("/")[0] == parent and p.rpartition("/")[2] != name
+            ]
+            entry["outcome"] = "merged under a new name"
+            if renamed:
+                entry["merged_as"] = renamed
+        conflicts.append(entry)
+
+    missing = [
```
healkeiser: **`missing` excludes anything in `existed`, so a requested path that exists locally but is absent from the merged file is never reported as not_found_in_file and is instead falsely reported as a rename conflict.**

The scene has /obj/null1; other.hip has no /obj/null1. merge_hip(f, ['/obj/null1']) returns merged_nodes [], not_found_in_file [], and conflicts [{'requested': '/obj/null1', 'existed': True, 'outcome': 'merged under a new name'}] - the caller is told the node arrived under a new name when nothing arrived at all. Confirmed by running the handler. The `conflicts` outcome should be derived from whether anything actually appeared, not from the pre-merge `existed` set alone.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
     }
 
 
+###### scene.merge_hip
+
+
+def _top_level_nodes() -> dict[str, set]:
```
healkeiser: **`_top_level_nodes()` only snapshots depth 2 (children of root contexts), so a node merged into an existing container is never listed in merged_nodes even though the `<path> <path>/*` pattern deliberately reaches deeper.**

merge_hip(f, ['/obj/geo1/sphere1']) where /obj/geo1 already exists locally. /obj/geo1 is unchanged at depth 2, so after - before is empty: merged_nodes [], merged_count 0, and the caller concludes nothing was merged although sphere1 is now in the scene. The container case the PR was written for (`/obj/geo1 /obj/geo1/*`) has the same hole: the container shows up but none of its children do. Diff the requested parents' allSubChildren() (or recurse) instead of only root-context children.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+    existed = [path for path in requested if hou.node(path) is not None]
+    warnings: list[str] = []
+    try:
+        hou.hipFile.merge(
+            file_path,
+            node_pattern=pattern,
+            overwrite_on_conflict=bool(overwrite_on_conflict),
+            ignore_load_warnings=False,
+        )
+    except hou.LoadWarning as warning:
```
healkeiser, line 219: **hou.LoadWarning is caught and discarded into a `warnings` list while `success` stays unconditionally True, so a merge that loaded nothing still reports success.**

The file is corrupt or every requested parent is missing; Houdini raises LoadWarning, the handler swallows it, and the result is {'success': True, 'merged_count': 0, 'warnings': [...]}. An agent that branches on `success` (the first field in the dict) proceeds as if the nodes are present. Set success False, or add an explicit `merged_anything` flag, when warnings exist and merged is empty.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+        overwrite_on_conflict: Overwrite same-named nodes instead of
+            renaming the merged copy.
+    """
+    require_inside_project_root(file_path, "hip file")
+    if not os.path.isfile(file_path):
+        raise FileNotFoundError(f"File not found: {file_path}")
+    if node_paths is not None and not isinstance(node_paths, (list, tuple)):
+        raise ValueError("node_paths must be a list of absolute node paths.")
+    requested = [str(p).rstrip("/") for p in (node_paths or [])]
+    for path in requested:
```
healkeiser, line 179: **The only validation is `startswith("/")`, so Houdini pattern metacharacters pass through into node_pattern and then corrupt the report.**

merge_hip(f, ['/obj/geo*']) is accepted. The pattern becomes `/obj/geo* /obj/geo*/*` and Houdini merges several nodes, but `hou.node('/obj/geo*')` is None both before and after, so the literal string lands in not_found_in_file with the "produced nothing" note while merged_nodes lists the nodes that did arrive - a self-contradicting result. Either reject `*?^%` in node_paths or document them and skip the per-path existence bookkeeping for patterned entries.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+            file_path,
+            node_pattern=pattern,
+            overwrite_on_conflict=bool(overwrite_on_conflict),
+            ignore_load_warnings=False,
+        )
+    except hou.LoadWarning as warning:
+        warnings = [line.strip() for line in str(warning).splitlines() if line.strip()]
+    after = _top_level_nodes()
+
+    merged: list[str] = []
```
healkeiser: **With overwrite_on_conflict=True the overwritten node keeps its path, so it never appears in the after-minus-before diff and merged_nodes/merged_count report zero for a successful overwrite.**

merge_hip(f, ['/obj/null1'], overwrite_on_conflict=True) on an existing /obj/null1 returns merged_nodes [] and merged_count 0, contradicting the tool docstring's "`merged_nodes` lists what arrived". tests/test_merge_hip.py::test_overwrite_is_reported_in_place bakes this in rather than catching it. Either include overwritten paths in merged_nodes or say in the docstring that overwrites appear only under `conflicts`.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+    existing node is overwritten in place; `conflicts` names both cases.
+
+    Args:
+        file_path: The .hip file to merge from (inside the project root).
+        node_paths: Absolute node paths to merge; default everything.
+        overwrite_on_conflict: Overwrite same-named nodes instead of
+            renaming the merged copy.
+    """
+    require_inside_project_root(file_path, "hip file")
+    if not os.path.isfile(file_path):
```
healkeiser: **os.path.isfile runs on the raw path while require_inside_project_root validated the hou.text.expandString-expanded one, so a Houdini variable path is accepted by the security check and then rejected as missing.**

merge_hip('$HIP/lookdev/other.hip'). require_inside_project_root expands and passes; os.path.isfile('$HIP/lookdev/other.hip') is False on every OS, so the call dies with FileNotFoundError even though hou.hipFile.merge would have resolved it fine. Copied verbatim from load_scene (line 158), so the fix belongs in a shared prologue: check isfile on hou.text.expandString(file_path).

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+    snapshot: dict[str, set] = {}
+    root = hou.node("/")
+    if root is None:
+        return snapshot
+    for context in root.children():
+        with contextlib.suppress(Exception):
+            snapshot[context.path()] = {child.path() for child in context.children()}
+    return snapshot
+
+
```
healkeiser, line 169: **Altitude/reuse: merge_hip is a strict superset of load_scene(merge=True) and duplicates its whole prologue, adding a 200th tool where two optional parameters on load_scene would do.**

load_scene(merge=True) and merge_hip(file_path) now do the same thing with different reporting and different warning handling, so every future fix (the $HIP expansion above, the LoadWarning handling, the project-root check) has to be made twice and will drift. The smaller change is node_paths/overwrite_on_conflict parameters on load_scene's merge branch and no new tool; it also avoids the tool-count/required_commands.json churn the PR body flags.

tests/test_merge_hip.py
```diff
@@ -0,0 +1,120 @@
...
+        assert hou.hipFile.merge.call_args.kwargs["overwrite_on_conflict"] is True
+        assert result["conflicts"][0]["outcome"] == "overwritten in place"
+
+    def test_load_warnings_come_back_as_data(self, monkeypatch):
+        hou = self._scene(monkeypatch, {"/obj": set()}, {"/obj": set()})
+
+        class LoadWarning(Exception):
+            pass
+
+        hou.LoadWarning = LoadWarning
```
healkeiser: **Test state is assigned directly onto the shared sys.modules['hou'] MagicMock instead of via monkeypatch, so it is never restored and leaks into every test module that imports hou afterwards.**

test_load_warnings_come_back_as_data sets `hou.LoadWarning = LoadWarning` (a locally defined Exception subclass) and `_scene` sets hipFile.path/hasUnsavedChanges return values; none are undone. Any later test that asserts on hou.LoadWarning being a MagicMock, or on hou.hipFile.path() returning a fresh Mock, now sees this module's values. The suite passes today only because of collection order (-p no:randomly); with pytest-randomly enabled (it is installed - the warnings summary shows the plugin stack) the order can flip. Use monkeypatch.setattr for all three.

tests/test_merge_hip.py
```diff
@@ -0,0 +1,120 @@
...
+import fxhoudinimcp_server.handlers.scene_handlers as scene  # noqa: E402
+
+
+class TestMergeHip:
+    def _scene(self, monkeypatch, before, after, existing=()):
+        hou = scene.hou
+        snapshots = iter([before, after])
+        monkeypatch.setattr(scene, "_top_level_nodes", lambda: next(snapshots))
+        monkeypatch.setattr(scene, "require_inside_project_root", lambda p, w: p)
+        monkeypatch.setattr(scene.os.path, "isfile", lambda p: True)
```
healkeiser: **`monkeypatch.setattr(scene.os.path, "isfile", ...)` patches the real stdlib os.path module globally, not a module-local alias, for the duration of every test in the class.**

scene.os is the stdlib `os` module, so os.path.isfile returns True for every path in the process while the test runs. If an assertion in these tests fails, pytest's traceback/assertion-rewrite machinery consults os.path during reporting and can misbehave or mask the real failure. Patch `scene.os.path.isfile` through a module-level indirection, or monkeypatch require_inside_project_root plus a narrow isfile that only special-cases the test path.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+            ]
+            entry["outcome"] = "merged under a new name"
+            if renamed:
+                entry["merged_as"] = renamed
+        conflicts.append(entry)
+
+    missing = [
+        path
+        for path in requested
+        if path not in existed and hou.node(path) is None and not any(m == path for m in merged)
```
healkeiser: **The third clause of the `missing` filter is dead code.**

If `path not in existed` (absent before the merge) and `hou.node(path) is None` (absent after it), the path cannot be in `merged` - merged only holds paths of nodes that exist. `not any(m == path for m in merged)` is therefore always True and never affects the result; it also reads as an O(n) scan where `path not in merged` would say the same thing. Drop it.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
```diff
@@ -172,6 +172,112 @@ def load_scene(file_path: str, merge: bool = False) -> dict:
...
+            overwrite_on_conflict=bool(overwrite_on_conflict),
+            ignore_load_warnings=False,
+        )
+    except hou.LoadWarning as warning:
+        warnings = [line.strip() for line in str(warning).splitlines() if line.strip()]
+    after = _top_level_nodes()
+
+    merged: list[str] = []
+    for context, paths in after.items():
+        merged.extend(sorted(paths - before.get(context, set())))
```
healkeiser: **Redundant work and a constant field in the result: the per-context `sorted()` is thrown away by the later `merged.sort()`, and `"existed": True` is the same for every conflicts entry.**

merged.extend(sorted(...)) sorts each context's delta and then merged.sort() re-sorts the whole list, so the inner sort is pure waste on a large merge. Separately, conflicts is built only from `existed`, so `entry['existed']` is True on every element and carries no information for the caller. Use `merged.extend(paths - before.get(context, set()))` plus one final sort, and drop the constant key.

python/fxhoudinimcp/tools/scene.py
```diff
@@ -6,6 +6,9 @@
 
 from __future__ import annotations
 
+# Built-in
+from typing import Any
+
 # Built-in
```
healkeiser: **The new import leaves a duplicated, now-orphaned `# Built-in` section comment above `# Third-party`.**

Lines 9-13 read `# Built-in` / `from typing import Any` / blank / `# Built-in` / `# Third-party`, so the file has two Built-in headers and the second labels nothing. Every sibling module (tools/nodes.py, tools/graph.py, ...) uses a single `# Built-in` then `# Third-party`. Delete line 12.

healkeiser (commented):
### Code review

15 findings. 14 are inline below. 1 land outside this diff and are listed here:

- `houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py`:155 - Altitude: the LoadWarning-as-data fix was added only to the new sibling; load_scene's own merge and load calls still let hou.LoadWarning escape as an exception.

  The PR description claims warnings come back as data "as load_scene already does", but load_scene calls hou.hipFile.merge(file_path) and hou.hipFile.load(..., suppress_save_prompt=True) with the default ignore_load_warnings=False and no try/except, so any hip with a missing asset still raises out of load_scene and the caller gets an error instead of a loaded scene plus warnings. The root-cause change is one shared `_merge_with_warnings` / `_load_with_warnings` helper used by both, not a second code path.


houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: an empty list merges nothing and says so; omitting node_paths merges everything.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: collisions come from `hou.hipFile.collisionNodesIfMerged` before the merge, and a renumbered copy is matched by Houdini's stem (`null1` -> `null2`) among the new children of the requested parent, excluding other requested paths; each copy is claimed once. Live test `test_named_nodes_arrive_and_a_collision_is_renumbered` merges `/obj/null1` + `/obj/building` and gets `merged_as: /obj/null2` only.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: conflicts are what `collisionNodesIfMerged` reports, not what existed locally; a path that brought nothing is `not_found_in_file` (live test `test_a_node_absent_from_the_file_is_not_found_and_not_success`).

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Confirmed live: `/obj/geo1/sphere1` lands inside an existing `/obj/geo1`. Fixed in 8658b34: the snapshot watches the parent of each requested path, so it is reported (live test `test_a_node_merged_into_an_existing_container_is_reported`). Children that come inside a merged container are counted in `merged_descendant_count` rather than listed.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: `success` is False when named nodes were asked for and none arrived, or when a whole-file merge brought nothing but warnings.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: `*?^[]` and spaces in node_paths are refused with a message; node_paths are paths, and omitting it merges everything.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34: an overwritten node is listed in `merged_nodes` and in `conflicts` as "overwritten in place" (checked live).

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Fixed in 8658b34, in the shared path: the existence check runs on `hou.text.expandString(file_path)`, for load and merge.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Agreed and done in 8658b34: `merge_hip` is gone; `load_scene` takes `node_paths` and `overwrite_on_conflict` with `merge=True` (sent only when used, so an older plugin still loads). No tool-count change any more.

tests/test_merge_hip.py
d-potrivaev: Fixed in 8658b34: every `hou` attribute is set through monkeypatch (`tests/test_load_scene_merge.py`).

tests/test_merge_hip.py
d-potrivaev: Fixed in 8658b34: the tests write a real file under `tmp_path`; nothing in `os.path` is patched.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Gone with the rewrite in 8658b34: attribution is per requested path.

houdini/scripts/python/fxhoudinimcp_server/handlers/scene_handlers.py
d-potrivaev: Gone in 8658b34: one sort, and conflict entries carry only `requested`, `outcome` and `merged_as`.

python/fxhoudinimcp/tools/scene.py
d-potrivaev: Fixed in 8658b34.

d-potrivaev:
Thanks for the review. All 15 findings are answered in 8658b34, one commit on top of the branch: 15 addressed as suggested. Each inline thread has its answer.

Findings outside the diff:
- **load_scene's own LoadWarning (scene_handlers.py:155).** Fixed in 8658b34: load and merge share the handling; warnings come back in `warnings` and the message counts them, the file stays loaded.

Verified: ruff check + format, `pytest -m "not integration"` (581 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_scene_live.py::TestLoadSceneMerge) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: all fixed, unit tests pass (581).

PR #82, opened 2026-09-16
feat(lops): get_usd_bound_material resolves the material a prim renders with and where the binding comes from

d-potrivaev:
## What this changes

get_usd_materials reads each prim's direct binding only. A prim that
inherits its material from an ancestor, or receives it through a
collection, is listed as unbound there -- while Karma, resolving the binding
with ComputeBoundMaterial, shades it. Finding out which material a prim
actually renders with, and why, meant pxr from execute_python.

get_usd_bound_material takes a list of prim paths and a purpose (full by default --
what Karma renders, falling back to an all-purpose binding -- preview, or all), resolves
them in one ComputeBoundMaterials call, and names the source: `direct` on the prim,
`inherited` with the ancestor that carries the binding, or `collection` with the
collection's path, plus the binding strength and the prim's direct binding for the same
purpose. A binding to a material prim that does not exist is reported in
`missing_material`, not as unbound; a prim the stage does not have answers in its own row.

get_usd_materials, the root cause, keeps `bound_to` (where a binding is authored) and adds
`rendered_on` / `rendered_on_count`: the gprims that resolve to each material for purpose
full. The Solaris guide routes "which material does this prim render with" to the new tool.

The tool count (server_instructions.md, README, docs) and required_commands.json overlap with
the neighbouring tool PRs; I'll rebase and recount after each merge.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 577 passed (`tests/test_usd_bound_material.py` and the updated neighbours)

Updated after review in 45937b4.


houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+
+    get_usd_materials lists direct bindings only; a prim bound through its
+    parent, or through a collection, showed up unbound. This resolves the
+    binding the way the renderer does (ComputeBoundMaterial) and names the
+    source: direct, inherited from which ancestor, or which collection.
+    Batched: pass every prim of interest in one call.
+    """
+    _require_pxr()
+    stage = _get_lop_stage(node_path)
+    token_name = _BINDING_PURPOSES.get(str(purpose).lower())
```
healkeiser, line 851: **The default purpose="all" maps to the allPurpose token, which is NOT what a renderer resolves, so the tool's headline claim ("the material each prim actually renders with") is wrong by default whenever a full- or preview-purpose binding exists.**

Verified on Houdini 22.0.368: /grp carries both an allPurpose binding to /materials/srf and a full-purpose binding to /materials/full_only. ComputeBoundMaterial(allPurpose) -> /materials/srf; ComputeBoundMaterial(full) -> /materials/full_only. Karma renders with the full purpose, so get_usd_bound_material(prim_paths=["/grp/sph1"]) with the default answers /materials/srf while the frame shades with /materials/full_only. Note allPurpose is the *narrowest* query, not a union: the name "all" invites exactly the opposite reading. ComputeBoundMaterial(full) already falls back to allPurpose internally, so "full" is the honest default.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+    binding_prim = None
+    with contextlib.suppress(Exception):
+        binding_prim = str(rel.GetPrim().GetPath())
+        source["binding_prim"] = binding_prim
+    name = ""
+    with contextlib.suppress(Exception):
+        name = rel.GetName()
+    if ":collection:" in name:
+        source["kind"] = "collection"
+        source["collection"] = name.split(":collection:", 1)[1]
```
healkeiser: **The collection name is parsed by splitting on ":collection:", which leaves the material purpose glued to the front for purpose-specific collection bindings.**

Verified: binding collection "furn" on /set for the preview purpose produces the relationship name `material:binding:collection:preview:furn`. `name.split(":collection:", 1)[1]` yields `"preview:furn"`, so the reply names a collection that does not exist on the stage and any follow-up lookup of it fails. USD already ships the right helper: `UsdShade.MaterialBindingAPI.CollectionBinding(rel).GetCollectionPath()` returns `/set.collection:furn` (and `CollectionBinding.IsCollectionBindingRel(rel)` replaces the substring test).

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+    for raw in prim_paths:
+        prim_path = str(raw)
+        prim = stage.GetPrimAtPath(prim_path)
+        if not prim or not prim.IsValid():
+            bindings.append({"prim": prim_path, "error": "prim not found on this stage"})
+            continue
+        api = UsdShade.MaterialBindingAPI(prim)
+        material, rel = api.ComputeBoundMaterial(token)
+        entry: dict[str, Any] = {"prim": prim_path, "material": None}
+        if material and material.GetPrim().IsValid():
```
healkeiser: **`source` is built only when the material resolves, so a binding that points at a missing material prim is reported as plain "unbound" - discarding the relationship ComputeBoundMaterial did return.**

Verified: /grp has `material:binding` -> /materials/missing (no such prim) and /grp/sph1 inherits it. ComputeBoundMaterial returns an invalid material but a *valid* rel named `material:binding` on /grp. The handler drops `rel` and emits `{"prim": "/grp/sph1", "material": null, "direct_binding": null}` - byte-identical to a prim with no binding at all. That is the exact "my prim renders grey, why" case the tool exists to answer. Build `source` whenever `rel` is valid, regardless of the material.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+            entry["material"] = str(material.GetPath())
+            entry["source"] = _binding_source(prim_path, rel)
+        with contextlib.suppress(Exception):
+            direct = str(api.GetDirectBinding().GetMaterialPath())
```
healkeiser: **`direct_binding` calls `GetDirectBinding()` with no argument, which always reads the allPurpose binding, so it contradicts the resolved row whenever purpose is "full" or "preview".**

Verified: with a prim whose only direct binding is a full-purpose one, purpose="full" produces `material: /materials/full_only`, `source.kind: "direct"` (binding_prim == the prim) and `direct_binding: null` in the same row - self-contradictory. With both an allPurpose and a full binding on /grp, purpose="full" reports `material: /materials/full_only` next to `direct_binding: /materials/srf`, which reads as "an ancestor overrode this prim" when nothing did. `api.GetDirectBinding(token)` is the fix; `GetDirectBinding('full')` returns /materials/full_only.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+
+    bindings: list[dict[str, Any]] = []
+    for raw in prim_paths:
+        prim_path = str(raw)
+        prim = stage.GetPrimAtPath(prim_path)
+        if not prim or not prim.IsValid():
+            bindings.append({"prim": prim_path, "error": "prim not found on this stage"})
+            continue
+        api = UsdShade.MaterialBindingAPI(prim)
+        material, rel = api.ComputeBoundMaterial(token)
```
healkeiser: **The docstring sells the tool as batched but the loop calls the per-prim ComputeBoundMaterial with no shared cache, re-walking ancestors and re-evaluating every collection membership query once per prim.**

N prims under a set with collection bindings: each iteration re-resolves the whole ancestor chain and rebuilds the collection query from scratch, so cost is O(N x depth x collections) instead of O(N + depth). `UsdShade.MaterialBindingAPI.ComputeBoundMaterials(prims, purpose)` exists in Houdini 22 (verified: returns `(materials, rels)`, with `None` entries for unbound prims) and does the whole list in one pass with shared binding and collection caches. On a shot with a few hundred prims per call this is the dominant cost of the tool that advertises "pass every prim of interest in one call".

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+        prim_paths = [prim_paths]
+    if not isinstance(prim_paths, (list, tuple)) or not prim_paths:
+        raise ValueError("prim_paths must be a prim path or a non-empty list of them.")
+
+    bindings: list[dict[str, Any]] = []
+    for raw in prim_paths:
+        prim_path = str(raw)
+        prim = stage.GetPrimAtPath(prim_path)
+        if not prim or not prim.IsValid():
+            bindings.append({"prim": prim_path, "error": "prim not found on this stage"})
```
healkeiser: **Error rows have a different shape from success rows: they carry no "material" key at all, and the summary counters cannot distinguish "not found" from "unbound".**

A caller that iterates `for b in reply["bindings"]: b["material"]` raises KeyError the moment one path is mistyped, even though every other row is guaranteed to have the key. Likewise `count - bound` reads as "unbound prims" but silently includes prims the stage does not have: passing ["/a", "/typo"] where /a is bound gives count=2, bound=1, suggesting one unbound prim when there are none. Add `"material": None` to the error row, or an `errors` counter.

tests/test_usd_bound_material.py
```diff
@@ -0,0 +1,163 @@
...
+        api_obj.GetDirectBinding.return_value.GetMaterialPath.return_value = _Path(
+            (direct or {}).get(path, "")
+        )
+        api_for[path] = api_obj
+        return api_obj
+
+    usd_shade = MagicMock()
+    usd_shade.MaterialBindingAPI.side_effect = api
+    usd_shade.MaterialBindingAPI.GetMaterialBindingStrength.return_value = "weakerThanDescendants"
+    usd_shade.Tokens.allPurpose = "allPurpose"
```
healkeiser: **The fake gives `UsdShade.Tokens.allPurpose` a truthy value that the real token does not have, and no test exercises the default purpose at all.**

Verified: the real `UsdShade.Tokens.allPurpose` is the empty string `''`. The fake sets `"allPurpose"`. Only `test_purpose_is_passed_as_the_usd_token` asserts a token reaches ComputeBoundMaterial, and it uses "preview"; the default path is never asserted. Any future truthiness guard on the token (`if token:`, `token or default`, a `.get(purpose) or fallback`) would pass the whole suite and silently change the default query live. Set the fake to `""` and add an assertion that the default calls `ComputeBoundMaterial("")`.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
```diff
@@ -760,6 +760,87 @@ def _get_usd_materials(*, node_path: str) -> dict[str, Any]:
...
+) -> dict[str, Any]:
+    """The material each prim actually renders with, and why.
+
+    get_usd_materials lists direct bindings only; a prim bound through its
+    parent, or through a collection, showed up unbound. This resolves the
+    binding the way the renderer does (ComputeBoundMaterial) and names the
+    source: direct, inherited from which ancestor, or which collection.
+    Batched: pass every prim of interest in one call.
+    """
+    _require_pxr()
```
healkeiser: **`_require_pxr()` is redundant: the very next line's `_get_lop_stage` calls it first thing.**

Dead line. `_get_lop_stage` (line 61) opens with `_require_pxr()`, and none of the other twenty-odd handlers in this file call both. Costs nothing at runtime but invites the reader to assume `_get_lop_stage` does not guard pxr, which is how the second copy appears in the next handler.

tests/test_usd_bound_material.py
```diff
@@ -0,0 +1,163 @@
...
+
+sys.modules.setdefault("hou", MagicMock())
+sys.modules.setdefault("hdefereval", MagicMock())
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "houdini", "scripts", "python"))
+
+# Internal
+import fxhoudinimcp_server.handlers.lops_handlers as lops  # noqa: E402
+
+
+class _Path:
```
healkeiser: **The `_Path` and `_prim` USD fakes are re-declared here, near-identical to the pair already used against the same module in tests/test_list_usd_prims_root.py.**

`tests/test_list_usd_prims_root.py:25` defines `class _Path` and `:43` defines `def _prim(path)` for `lops_handlers` with the same shape. Two divergent copies of the same USD stub now have to be kept in step; a third lops test will make three. A conftest fixture (or importing the existing pair) is the one-line alternative.

healkeiser (commented):
### Code review

11 findings. 9 are inline below. 2 land outside this diff and are listed here:

- `houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py`:742 - The root cause is left in place: `_get_usd_materials` still builds `bound_to` from `GetDirectBinding()` only, and nothing points a reader of that tool at the new one.

  The PR's own premise is that get_usd_materials under-reports bindings, yet its binding loop (which already traverses every prim on the stage) is untouched, its docstring gains no cross-reference, and `python/fxhoudinimcp/prompts/markdown/workflows/solaris.md:44` still instructs the agent to use `get_usd_materials` "to verify hierarchy, bindings and the layer stack". An agent following the Solaris guide sees `bound_to: ["/grp"]`, concludes /grp/sph1 is unbound, and never learns get_usd_bound_material exists. Fixing the loop (or adding a computed `bound_to`) fixes every caller at once instead of adding a second tool nobody is routed to.

- `docs/how-to/tools.md`:5 - The PR bumps the tool count only in server_instructions.md, leaving four doc locations claiming 199 tools and an 18-tool LOPs section.

  Before this PR, server_instructions.md, docs/how-to/tools.md:5, docs/how-to/tools.md:69 ("LOPs/USD (18 tools)"), docs/how-to/index.md:6 and docs/index.md:7/:52 all agreed on 199. After it, the server advertises 200 while every published doc page still says 199 and LOPs still says 18. `tests/test_prompts_resources.py::TestInstructionHeaderCounts` guards only server_instructions.md, so nothing catches the drift and the published site is silently wrong.


houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Confirmed on 22.0.429. Fixed in 45937b4: the default is `full` (which falls back to allPurpose by itself); `all` stays available and the docstring says it is the narrowest query.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Fixed in 45937b4: `CollectionBinding.IsCollectionBindingRel` + `GetCollectionPath()` → `/set.collection:furn` (checked live with a preview-purpose collection binding).

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Fixed in 45937b4: when the relationship is valid its `source` is reported and `missing_material` lists the target (`/materials/missing` in the live test); the summary counts them.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Fixed in 45937b4: `GetDirectBinding(token)`, the same purpose as the resolution.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Fixed in 45937b4: one `ComputeBoundMaterials(prims, purpose)` call for the whole list.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Fixed in 45937b4: every row has `material`; summary has `bound`, `missing_material` and `not_found`.

tests/test_usd_bound_material.py
d-potrivaev: Fixed in 45937b4: the fake token is `""`, and tests assert the default call uses `full` and `purpose="all"` passes `""`.

houdini/scripts/python/fxhoudinimcp_server/handlers/lops_handlers.py
d-potrivaev: Removed in 45937b4.

tests/test_usd_bound_material.py
d-potrivaev: Fixed in 45937b4: `tests/_usd_fakes.py`, used by this test and `test_list_usd_prims_root.py`.

d-potrivaev:
Thanks for the review. All 11 findings are answered in 45937b4, one commit on top of the branch: 11 addressed as suggested. Each inline thread has its answer.

Findings outside the diff:
- **The root cause in get_usd_materials (lops_handlers.py:742).** Agreed that the root cause is in get_usd_materials, and 45937b4 fixes it there: `bound_to` keeps meaning "where a binding is authored", and `rendered_on` / `rendered_on_count` list the gprims that resolve to each material for purpose `full` (one ComputeBoundMaterials batch over the traversal that already runs). The Solaris guide and the docstring route "which material does this prim render with" to get_usd_bound_material. I kept the tool: it answers a per-prim "why" — direct, inherited from which ancestor, which collection, missing material — that a material list cannot. If you would rather have only the get_usd_materials fix, I will drop the new tool from the PR.
- **Tool count in README/docs.** Fixed in 45937b4: 200 tools, LOPs/USD 19. Also added `prim_paths` to STAND_INS — the path-validation test failed on the new command.

Verified: ruff check + format, `pytest -m "not integration"` (577 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_usd_bindings_live.py) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: all fixed, unit tests pass (577).

PR #83, opened 2026-09-16
feat(hda): set_hda_interface and edit_hda_interface author and edit an asset's Type Properties interface

d-potrivaev:
## What this changes

create_spare_parameter adds parameters to a node instance and never reaches
the type, so an asset's own interface — tab folders, strict ranges, ordered
menus, Hide/Disable When — had no verb, and once written nothing could edit
it: a session that rebuilt a Controls tab (remove ~40 parameters, insert
~30, a hidden parameter with a default expression, a button with a Python
callback, Base1 naming on every 2-component template) went to execute_python
five times.

set_hda_interface(node_path, parameters, replace=False) writes the
definition's parameter interface from a declarative spec, validated before
the write (duplicate names, unknown types, a menu without items, a default
outside its items), and reports the interface read back off the definition.
Houdini renames a tab folder appended next to an existing tab group (it joins
the group and takes its naming series, renumbering the whole series along
the way — "controls" landed as "stdswitcher4_3" live); the reply names that
in renamed_by_houdini rather than reporting the parameter lost.

edit_hda_interface(node_path, ops, dry_run=False) takes a list of operations
over the definition's ParmTemplateGroup — insert (after / before /
in_folder), remove, hide / show, replace, modify (label, help, default,
default_expression, min/max, conditionals, menu items, callback, naming
scheme, rename, tags), move — applies them all to a copy, and writes once;
a failing op leaves the definition untouched. Two things measured live on
22.0.429 shape it:

* Houdini refuses the whole group over a component-name collision and says
  only "The attempted operation failed": a 2-float `fh_range` under Base1
  makes `fh_range1`/`fh_range2`, and a template called `fh_range2` next to it
  cannot be written. The names each template will create are predicted from
  its naming scheme and the collision is reported by name before the write.
* Built-in parameters of the node type cannot be removed from an asset's
  interface: remove("s") and removing the Subnet folder go through in the
  group, and after setParmTemplateGroup Houdini puts them back at the top
  level without a word. The reply lists them in reinstated_by_houdini and
  says to hide them instead.

The spec grows the types button (with a callback, Python by default),
separator, label, vector, color, file, oppath and the fields naming_scheme,
default_expression, hidden, join_with_next, callback, tags; folder_type
"multiparm" builds a multiparm block whose children are named "item#".

Verified live on Houdini 22.0.429 against an Object subnet turned into an
asset: seven ops in one call — fh_range1/fh_range2 (Base1), clrr/clrg/clrb,
a hidden int whose default expression ch('tx')*2 evaluates to 10 at tx=5, a
button whose callback sets the node comment, item1/item2 from the multiparm,
`s` reported reinstated, `r` stored hidden; get_parm_template_tree shows the
five parameters in insertion order with their DisableWhen; a second call
moves, renames and shows.

## How it was verified

- [x] `pytest tests -m "not integration" -q` — 615 passed (`tests/test_hda_interface.py` and the updated neighbours)

Updated after review in b7c8bc3.


healkeiser (commented):
### Code review

14 findings. 14 are inline below.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+        return record
+    if kind in ("remove", "hide", "show"):
+        target, target_kind = _find_target(
+            group, op.get("name") or op.get("folder"), f"op #{index}"
+        )
+        record["target"] = target.name()
+        record["target_kind"] = target_kind
+        if kind == "remove":
+            group.remove(target)
+            existing.discard(target.name())
```
healkeiser: **Removing a folder only discards the folder's own name from `existing`, not its children, so re-inserting a parameter that lived in that folder is falsely rejected as already existing.**

The PR's own headline workflow: ops = [{"op":"remove","name":"Controls"}, {"op":"insert","spec":{"name":"stud_count",...}}]. hou docs (ParmTemplateGroup.remove): "if you remove a folder, all of the parm templates in the folder are also removed" - but line 1118 runs only `existing.discard(target.name())` ("Controls"). Op #1 then hits the guard at line 1100-1105 and raises "a parameter named 'stud_count' already exists; use 'replace' or 'modify' for it", even though nothing in the group holds that name any more. Rebuilding a tab (remove ~40, insert ~30) is impossible in one atomic call. Same hole in the `replace` branch at line 1137.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+        group.clear()
+
+    seen = set()
+    requested = []
+    for spec in parameters:
+        template = _build_parm_template(spec, seen)
+        group.append(template)
+        requested.append(template.name())
+
+    definition.setParmTemplateGroup(group)
```
healkeiser: **set_hda_interface appends blindly: it never checks the requested names against the existing group, never runs `_component_collisions`, and does not wrap `setParmTemplateGroup` - so it produces exactly the bare "The attempted operation failed" the PR was written to eliminate.**

The `seen` set at line 771 only rejects duplicates *within one request*. Call set_hda_interface twice with the same spec (a retry, a re-run of a script), or add a 2-float `fh_range` under Base1 next to an existing `fh_range2`: `HDADefinition.setParmTemplateGroup(group, rename_conflicting_parms=False)` raises hou.OperationFailed (documented: "If False and there are parms with the same name, this method raises OperationFailed"). Line 778 has no try/except, so the caller gets a raw unhelpful traceback, while the sibling verb at line 1206-1216 and 1233-1238 gives a named collision and "Nothing was written." The pre-check and the guard belong in the shared write path, not in one of the two verbs.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+    # asset's own "stdswitcher"/"_1"/"_2" became "stdswitcher4"/"_1"/"_2" and
+    # the folder asked for as "controls" landed as "stdswitcher4_3"). So: try
+    # the requested name, then fall back to position — appended entries sit at
+    # the end, in the order they were appended.
+    stored = list(definition.parmTemplateGroup().parmTemplates())
+    matched = []
+    for offset, want in enumerate(requested):
+        hit = next((t for t in stored if t.name() == want), None)
+        if hit is None:
+            index = len(stored) - len(requested) + offset
```
healkeiser: **The positional fallback in set_hda_interface's readback can bind a requested name to an unrelated pre-existing template, reporting a bogus rename and hiding a genuinely lost parameter.**

`index = len(stored) - len(requested) + offset` assumes the appended entries are still the last N top-level entries. Both behaviours this PR itself documents break that: Houdini reinstates removed built-ins *at the top level* (comment at line 1242-1244) and renumbers tab series. Append one folder "controls" to an Object asset; Houdini renames it and re-adds `s` at the end -> stored = [..., stdswitcher4_3, s], index points at `s`, and the reply says {"requested": "controls", "stored_as": "s"} with `applied` describing the built-in `s`. Worse, if Houdini genuinely drops the template the fallback still finds *something*, so `not_found_after_write` is empty and real loss is reported as success.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+
+    stored_group = definition.parmTemplateGroup()
+    stored = {t.name(): t for t in _flatten_templates(stored_group.entries())}
+    # Built-in parameters of the node's own type (an Object's Transform,
+    # its Subnet folder) cannot be removed from a definition: Houdini puts
+    # them back, at the top level, and says nothing (measured live).
+    reinstated = sorted(name for name in result["removed"] if name in stored)
+    result["removed"] = [name for name in result["removed"] if name not in reinstated]
+    result["reinstated_by_houdini"] = reinstated
+    result["not_found_after_write"] = sorted(name for name in planned if name not in stored)
```
healkeiser: **edit_hda_interface has none of the tab-folder-rename handling set_hda_interface was built around, so an inserted folder Houdini renames is reported as lost.**

`stored` at line 1241 is keyed by name and `not_found_after_write` at line 1248 is a plain name lookup. Insert a folder named "controls" next to an existing tab group: Houdini stores it as "stdswitcher4_3" (the PR's own measured case). The reply then lists "controls" in `added`, in `not_found_after_write` and in `instance_parms_missing` with no `renamed_by_houdini` key and no note - the exact "reported the parameter lost while the interface was sitting right there" failure the other verb spends 25 lines avoiding.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+
+    applied = [_describe_template(hit) for _want, hit in matched if hit is not None]
+    renamed = [
+        {"requested": want, "stored_as": hit.name(), "label": hit.label()}
+        for want, hit in matched
+        if hit is not None and hit.name() != want
+    ]
+    lost = [want for want, hit in matched if hit is None]
+
+    node_parms = {parm.name() for parm in hou.node(node_path).parms()}
```
healkeiser: **set_hda_interface's instance-parm check compares template names against `node.parms()` only (no `parmTuples()`) and walks `children` exactly one level deep, so multi-component and nested parameters are always reported missing.**

The PR's own verified case: a 3-component color named `clr` creates `clrr`/`clrg`/`clrb` on the node, so `"clr" in node_parms` is False and `instance_parms_missing` reports `["clr"]` on a completely successful write. edit_hda_interface fixed this at line 1262 by also consulting `parmTuples()`; set_hda_interface was never updated. Separately, line 810 iterates `spec["children"]` once, so a folder inside a folder contributes the *inner folder's name* (never a parm) to expected_parms and its real children are skipped entirely - guaranteed false "missing" plus guaranteed silent omissions.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+def _component_collisions(group) -> list[dict]:
+    """Pairs of templates whose component names collide."""
+    owners: dict[str, str] = {}
+    collisions: list[dict] = []
+    for template in _flatten_templates(group.entries()):
+        # The tuple's own name is a name on the node as well: `fh_range2` the
+        # tuple collides with `fh_range2` the second component of `fh_range`.
+        for component in dict.fromkeys([*_component_names(template), template.name()]):
+            other = owners.get(component)
+            if other is not None and other != template.name():
```
healkeiser: **`_component_collisions` skips any pair where `other == template.name()`, so it structurally cannot detect the plain duplicate-name case - which is precisely the case HDADefinition.setParmTemplateGroup raises OperationFailed on; and `modify`/`new_name` and `replace` never check the new name against `existing` either.**

`if other is not None and other != template.name()` (needed because folders in one tab set legitimately share a name - hou docs: "folder parm templates in the same folder set will share the same name") means two distinct templates both named `size` are silently accepted. Concrete path: {"op":"rename","name":"width","new_name":"size"} where `size` already exists - line 1148-1151 just moves the name in `existing` with no conflict check, the collision scan is blind to it, and Houdini refuses the whole group with a bare OperationFailed. Same for `replace` (line 1138) and for two `add_multiparm` ops whose children are both the docstring-suggested `item#` (both produce `item1`, `item2`, ... on the node).

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+                )
+        names = record.get("names") or ([target] if target else [])
+        record["stored"] = [_describe_template(stored[n]) for n in names if n in stored]
+
+    node_parms = {parm.name() for parm in hou.node(node_path).parms()}
+    tuple_names = {t.name() for t in hou.node(node_path).parmTuples()}
+    result["instance_parms_present"] = sorted(
+        n for n in result["added"] if n in node_parms or n in tuple_names
+    )
+    result["instance_parms_missing"] = sorted(
```
healkeiser: **edit_hda_interface's `instance_parms_missing` is computed over `added`, which contains folder names and multiparm children - neither of which ever appears in node.parms()/parmTuples() - producing false alarms on every successful folder or multiparm insert.**

Insert a tab folder "controls" and a multiparm with a child "item#": `planned - before_names` includes both "controls" and "item#". A non-first tab folder has no parm tuple at all (hou docs: the folder-set tuple takes the name of the *first* folder in the set), and the node carries `item1`/`item2`, never `item#`. Both land in `instance_parms_missing`, telling the caller the write half-failed when it did not.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+            default_value=_component_defaults(spec, components, str, ""),
+            string_type=string_type,
+            naming_scheme=_naming_scheme(spec, hou.parmNamingScheme.Base1),
+            **kwargs,
+            **common,
+        )
+    elif kind == "button":
+        template = hou.ButtonParmTemplate(name, label, **common)
+    elif kind == "separator":
+        template = hou.SeparatorParmTemplate(name)
```
healkeiser: **The `separator` branch discards the whole `common` kwargs dict, so `hidden: true` on a separator spec is silently ignored even though hou.SeparatorParmTemplate accepts `is_hidden`.**

`hou.SeparatorParmTemplate(name)` - the documented signature is `SeparatorParmTemplate(self, name, is_hidden=False, tags={})`. A spec `{"name": "sep1", "type": "separator", "hidden": true}` validates, writes, and produces a visible separator; the reply's `applied`/`stored` entry (line 704-727) never reports `hidden`, so nothing tells the caller the field was dropped. Related dead code: line 559 filters `common` for `"help"`, but `_common_template_kwargs` (line 663-677) never puts `help` in the dict, so that key can never match.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+    suffixes = _COMPONENT_SUFFIXES.get(scheme, [str(i + 1) for i in range(count)])
+    return [f"{name}{suffixes[i] if i < len(suffixes) else i + 1}" for i in range(count)]
+
+
+def _flatten_templates(entries) -> list:
+    """Every template in *entries*, folders included, depth-first."""
+    flat = []
+    for entry in entries:
+        flat.append(entry)
+        if entry.type().name() == "Folder":
```
healkeiser: **`_flatten_templates` recurses into multiparm blocks, contradicting the guidance in hou's own ParmTemplateGroup docs, so multiparm children pollute `before_names`, `added`, `removed` and the collision scan.**

The check is `entry.type().name() == "Folder"`, but a multiparm block *is* a Folder-type template (hou docs: "folder parm templates are also used to represent multiparm blocks; the folder parm template's hou.folderType determines whether it is a folder or a multiparm block"). The docs' own `allParmTemplates` example guards with `isActualFolder()` for exactly this reason. Consequence: `item#` appears in `added`, a `remove` of a multiparm reports every `<child>#` in `removed`, and a folder-vs-multiparm `in_folder` target resolves to something `appendToFolder` treats differently. Use `entry.isActualFolder()`.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+    if info["type"] == "Folder":
+        info["children"] = [_describe_template(child) for child in template.parmTemplates()]
+    else:
+        for attr, key in (("minIsStrict", "min_strict"), ("maxIsStrict", "max_strict")):
+            with contextlib.suppress(Exception):
+                info[key] = bool(getattr(template, attr)())
+    return info
+
+
+def set_hda_interface(
```
healkeiser, line 754: **set_hda_interface is a strictly weaker duplicate of edit_hda_interface plus append-only insert ops; the two write paths have already drifted apart and every bug above lives in the copy that was not maintained.**

set_hda_interface == `edit_hda_interface(ops=[{"op":"insert","spec":s} for s in parameters])` with an optional `group.clear()` first. Because it is a second implementation, it lacks the collision pre-check, the try/except around the write, the `parmTuples()` readback, the `reinstated_by_houdini` reporting, `dry_run`, and the recursive children walk - four of the findings in this list exist only in this copy. Making it a thin wrapper that builds insert ops and calls the edit path is a smaller diff than fixing each divergence, and fixes them at the root.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+        raise ValueError(
+            f"{spec.get('name')!r}: {len(expressions)} default expression(s) for "
+            f"{components} component(s)."
+        )
+    language = _script_language(spec, "default_expression_language")
+    template.setDefaultExpression(tuple(expressions))
+    template.setDefaultExpressionLanguage(tuple([language] * components))
+
+
+def _describe_template(template) -> dict:
```
healkeiser, line 732: **`_describe_template` and `_build_parm_template` re-implement `parameter_handlers._template_to_dict` and `parameter_handlers._build_parm_template` under the same names, giving the codebase two incompatible parameter-spec dialects and a readback poorer than the tool the docstrings tell callers to use.**

`parameter_handlers.py:76 _template_to_dict` already reports name, label, type, num_components, is_hidden, default_value, min/max, min/max_is_strict, menu items and naming_scheme - and is what `get_parm_template_tree` returns. The new `_describe_template` reports only name/label/type/conditionals/min_strict/max_strict, so `applied`/`stored` omit the default value and the hidden state, i.e. the two things a caller most wants to confirm after a write. Meanwhile `parameter_handlers.py:614 _build_parm_template(spec)` speaks `parm_name`/`parm_type`/`default_value`/`min_val` while the new one speaks `name`/`type`/`default`/`min`, so create_spare_parameters and set_hda_interface take different JSON for the same concept. Both new functions also bypass the module's existing `_get_definition(node)` helper (line 33) to inline their own definition-is-None check.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+        "ops": records,
+        "added": sorted(planned - before_names),
+        "removed": sorted(before_names - planned),
+    }
+    if dry_run:
+        result["message"] = f"{len(records)} op(s) validated; nothing written."
+        return result
+
+    try:
+        definition.setParmTemplateGroup(group)
```
healkeiser, line 1251: **Both new verbs write the asset's .hda library file on disk with no `require_inside_project_root` guard, although README documents HDA libraries as covered by FXHOUDINIMCP_PROJECT_ROOT.**

`HDADefinition.setParmTemplateGroup` saves the definition to its hda file (and, with the default `create_backup=True`, drops a backup beside it). Point the tool at an instance of a studio asset installed from a shared library and the shared .hda is rewritten, wherever it lives. README.md:465 states "hip files, imports, exports and HDA libraries must live under this directory" and create_hda/install_hda/uninstall_hda all call `require_inside_project_root(file_path, "HDA file")`. Passing `definition.libraryFilePath()` through that same guard closes it. (update_hda and set_hda_section_content share the gap; this PR widens it.)

python/fxhoudinimcp/prompts/markdown/instructions/server_instructions.md
```diff
@@ -1,4 +1,4 @@
-Houdini MCP: 199 tools across 23 categories. Full rules: get_workflow_guide("discipline"), read once per session before building.
+Houdini MCP: 201 tools across 23 categories. Full rules: get_workflow_guide("discipline"), read once per session before building.
```
healkeiser, line 1: **Only the test-covered tool count was bumped; README.md and docs/index.md still advertise 199 tools and 11 HDA tools.**

`tests/test_prompts_resources.py:161` enforces the "N tools across M categories" sentence, so server_instructions.md was updated - but nothing checks the docs. After this PR `python/fxhoudinimcp/tools/hda.py` has 13 `@mcp.tool()` decorators while README.md:74 and docs/index.md:25 say `| **HDAs** | 11 |`, and "199 tools" survives at README.md:12, 57, 102, 594, 607 and docs/index.md:7, 52, 83. Note also that main already moved to 200 locally, so the hand-written 201 will be wrong the moment a sibling PR lands; the count wants generating, not typing.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
```diff
@@ -417,6 +418,863 @@ def create_hda(
...
+                raise ValueError(
+                    f"Default {default!r} of menu {name!r} is not one of its items {list(values)}."
+                )
+            default = values.index(default)
+        template = hou.MenuParmTemplate(
+            name,
+            label,
+            menu_items=values,
+            menu_labels=labels,
+            default_value=int(default),
```
healkeiser: **A menu's integer default is passed straight to Houdini with no range check, while the string form is validated - the PR body claims "a default outside its items" is validated.**

`{"name": "material", "type": "menu", "default": 7, "menu_items": ["plastic", "metal"]}` passes validation and stores `default_value=7` on a 2-item menu; the equivalent `"default": "gold"` is rejected at line 565-568. The guard is one line: reject when `not 0 <= int(default) < len(values)`. Same asymmetry in `_modify_template` at line 1021-1022.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: remove and replace free every name inside the target; live test removes "Controls" and re-inserts `stud_count` in one call.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3 by making it a wrapper over the edit path: names already in the interface are refused by name before the write (a retry is the live test), the collision scan runs, and the write is guarded.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: The positional fallback is gone in b7c8bc3. A planned folder missing by name is matched to an unclaimed stored folder by label (measured: the label survives the rename); anything unmatched is `not_found_after_write`.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3 in the shared readback. Measured live: the whole series is renumbered, built-ins included, so the reply separates `renamed_by_houdini` (folders the caller placed) from `renumbered_by_houdini` (folders that were already there).

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: value templates at any depth, checked against `parms()` and `parmTuples()` (a color `clr` is found; live test).

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: same-name templates collide unless both are folders of one tab set; rename, replace and insert check names against the interface first; multiparm children are scanned, so two `item#` blocks collide.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: folders and multiparm children are not expected on the instance.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: `SeparatorParmTemplate(name, is_hidden=...)` (checked live); the dead `help` key is gone from the label filter.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: multiparm children are left out (`isActualFolder()` is False for a multiparm block, measured) except in the collision scan.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Done in b7c8bc3: set_hda_interface builds insert ops (after `clear()` for replace) and calls the one write path; it gained `dry_run` on the way.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: the readback is `_template_to_dict` plus conditionals/children (default value and hidden state included), `_get_definition` is used, and specs accept create_spare_parameters' spelling (`parm_name`, `parm_type`, `default_value`, `min_val`, `max_val`). The richer spec (folders, conditionals, callbacks) stays in hda_handlers; happy to move create_spare_parameters onto it in a separate PR if you want one dialect.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3: the definition's library goes through `require_inside_project_root` before the write (update_hda and set_hda_section_content share the gap; I left them for a separate change).

python/fxhoudinimcp/prompts/markdown/instructions/server_instructions.md
d-potrivaev: Fixed in b7c8bc3: 201 tools, HDAs 13. On the count going stale when a sibling PR lands first: agreed it wants generating; not in this PR.

houdini/scripts/python/fxhoudinimcp_server/handlers/hda_handlers.py
d-potrivaev: Fixed in b7c8bc3 on build and on modify (live: "Default 7 of menu 'm' is out of range: it has 2 item(s), 0-1."). Also `ops` added to STAND_INS, and a bad node path is now named before a bad spec — the path-validation test failed on both commands.

d-potrivaev:
Thanks for the review. All 14 findings are answered in b7c8bc3, one commit on top of the branch: 14 addressed as suggested. Each inline thread has its answer.

Verified: ruff check + format, `pytest -m "not integration"` (615 passed), `gen_prompt_vocab.py --check`, and the integration suite in hython on Houdini 22.0.429 / Linux: the new live tests (test_hda_interface_live.py) pass, and the only failures are the 7 a clean `main` also has on Linux (`Q:/...` paths are plain writable directories there, the instruction-header parser tests, `layout_children` in path validation).

healkeiser (approved):
Re-verified every review thread against the head: findings are fixed or explicitly deferred, unit tests pass (615). One leftover, embedded HDAs hitting the project-root guard with the literal path "Embedded", will be fixed on main after merge.

